PR #465, opened 2014-02-05
Add a script to import & export db

dosomder:
I made a script for importing & exporting the nipap PostgreSQL database. Useful to back up the database daily using cronjob for example.


plajjan:
I closed this because there's too many things in here that don't resonate well with me. Like a bunch of SQL queries. I'm still worried about triggers and overall I think there are just better ways of dealing with it than this.

PR #522, opened 2014-07-18
User impersonation and vlan smart searching

edechaux:
These 4 commits reintroduce user impersonation for the nipap command line and add the possibility to smart search on the vlan field. Also includes a search command for the nipap tool in order to smart search with the command line.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -388,6 +388,17 @@ def list_prefix(arg, opts):
 
 
 
+def impersonate_user():
+    """ Provide a quick way to perform user impersonation
+        with the nipap_cli package.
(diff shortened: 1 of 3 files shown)


plajjan:
I've added #684 for the part about CLI output, so essentially, there are three parts to this:

* [ ] User impersonation (#540)
* [ ] CLI output (#684)
* [ ] Advanced search queries (#531)

plajjan:
And with the merge of PR #823, which closes #531, we should now fulfil all the requirements outlined in this PR, though with somewhat different solutions so I'm closing this.

plajjan:
That is, with the features of #531, you can now search for a specific VLAN using "vlan=100". It's the exactly the @ operator, but I think it's much better, since with this you can also do "vlan >= 100 and vlan < 200" to get all VLANs from 100 to 200, for example.

Issue #579, opened 2014-08-14
Make pynipap compatible with Python3


plajjan:

plajjan:
This is actually fixed with 601175b. Closing.

PR #594, opened 2014-08-28
Add compatibility for py3 to pynipap

dosomder:
Issue: https://github.com/SpriteLink/NIPAP/issues/579
Info taken from http://python3porting.com/noconv.html

Additional note: xmlrpc.client is a standard library of py3.

### DIFF
--- a/pynipap/pynipap.py
+++ b/pynipap/pynipap.py
@@ -100,7 +100,7 @@
 
         # Fetch VRF with ID 1 and print its name
         vrf = VRF.get(1)
-        print vrf.name
+        print(vrf.name)
 
     To list all objects each object has a :func:`list`-function. ::
 
@@ -109,7 +109,7 @@
 
         # print the name of the pools
         for p in pools:
-            print p.name
+            print(p.name)
 
     Each of the list functions can also take a `spec`-dict as a second
     argument. With the spec you can perform a simple search operation by
@@ -147,7 +147,12 @@
     Classes
     -------
 """
-import xmlrpclib
+try:
+    import xmlrpc.client as xmlrpclib
+except ImportError:
+    import xmlrpclib
+    int = long
+
 import logging
 
 __version__		= "0.26.4"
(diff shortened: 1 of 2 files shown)


pynipap/pynipap.py
```diff
@@ -147,7 +147,12 @@
     Classes
     -------
 """
-import xmlrpclib
+try:
```
plajjan: Explicitly check Python version (like performed in setup.py) instead of just blindly trying py3 import and failing over to py2 import.

pynipap/pynipap.py
```diff
@@ -147,7 +147,12 @@
...
+    import xmlrpc.client as xmlrpclib
+except ImportError:
+    import xmlrpclib
+    int = long
```
plajjan: Somewhat weird not being able to actually use int after this.. but I guess it's ok. I prefer py3 "correctness" in code (ie using int) than being closer to py2.

pynipap/pynipap.py
```diff
@@ -100,7 +100,7 @@
 
         # Fetch VRF with ID 1 and print its name
         vrf = VRF.get(1)
-        print vrf.name
+        print(vrf.name)
```
plajjan: Py2.7 works with this type of print function instead of the print statement but by also importing from __future__ import print_function I think we disable the old type of print statements and guarantee that all prints use the print() function available in py3.

pynipap/setup.py
```diff
@@ -26,7 +31,8 @@
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
         'Operating System :: POSIX :: Linux',
-        'Programming Language :: Python :: 2.6',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
 		'Topic :: Internet :: WWW/HTTP :: WSGI :: Middleware'
```
plajjan: Erroneous whitespace at start of line?

pynipap/pynipap.py
plajjan: This was a silly comment.. the print() line is in the docstrings at the top of the file. Naturally we don't have print statements in pynipap.. it's a lib after all ;)

Issue #604, opened 2014-09-04
Update to ldap3 library


dosomder:
The ldap library currently used for ldap authentication is not py3 compatible. There is a newer library called ldap3 (or python3-ldap) which is compatible to py3 and py2.6+. http://pythonhosted.org/python3-ldap/

PR #611, opened 2014-09-09
Replace ldap lib with ldap3

dosomder:
Replace ldap with ldap3 because ldap lib is not py3 compatible.
Functionality is the same.

https://github.com/SpriteLink/NIPAP/issues/604

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,8 @@ install:
     - wget -O - https://spritelink.github.io/NIPAP/nipap.gpg.key | sudo apt-key add -
     - sudo apt-get update -qq
     # install dependencies for installing & running nipap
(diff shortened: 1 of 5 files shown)


plajjan:
I'd like to get this integrated. The one remaining blocker was how to provide an ldap3 debian package. ldap3 packages will become available in jessie which is about to be released soon. Unfortunately, I don't think we can expect everyone to upgrade very soon so instead I've decided that we'll roll our own packages (python-ldap3 and python3-ldap3 respectively) in the meantime. Users of jessie or new versions of Ubuntu will rely on their own packages while users of older distros can use our packages.

There's some other work being done on our LDAP module (get AD working). That will most definitely collide with this patch but as this needs to be rebased already there is no way to avoid at least some work.

Would you mind have a go at this again? :)

plajjan:
@dosomder would you like to update this so that it applies cleanly on master?

dosomder:
Sorry I don't have time anymore to work on this project.

plajjan:
@dosomder ok, I understand. Thank you for your contributions thus far :) I will try to brush off this PR and ready it for merge.

Issue #605, opened 2014-09-04
Internal server error on ldap connection error


dosomder:
www-ui shows internal server error if it can not connect to ldap server. There is an AuthError in apache error.log

The www-ui should show a more specific error message.

plajjan:
This is fixed through the merge of #757.

PR #612, opened 2014-09-10
Show authentication error in web-ui

dosomder:
Instead of 'Internal Server error', the web-ui should show a better message. Currently this is only the case if the LDAP server is unavailable.

https://github.com/SpriteLink/NIPAP/issues/605

### DIFF
--- a/nipap-www/nipapwww/controllers/auth.py
+++ b/nipap-www/nipapwww/controllers/auth.py
@@ -37,8 +37,12 @@ class AuthController(BaseController):
         # Verify username and password.
         auth_fact = AuthFactory()
         auth = auth_fact.get_auth(request.params.get('username'), request.params.get('password'), 'nipap')
-        if not auth.authenticate():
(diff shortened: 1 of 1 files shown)


garberg:
Is it really a good idea to display the exception to the user? I guess there are cases when that'll give away too much information to a potentially malicious visitor. Wouldn't it be better to display a generic error message and log the exception (something like ```log.error("Authentication error: %s" % str(e))```) instead?

plajjan:
I agree with @garberg. We should always strive to not display raw exceptions to users and we should especially not display exceptions to users who haven't even logged in yet as that is a potential security risk.

Since @dosomder doesn't have time to work anymore, I suggest that @garberg rebases and fixes this up by printing a more general error message. I do agree with @dosomder's intent to differentiate between incorrect user credentials and pure system error, I just don't want to print as much information to non-logged-in users.

plajjan:
Closing as #757 implemented these changes.

PR #678, opened 2015-01-07
Fix pynipap status value

### LINKED ISSUE #668: Not specifying status results in error
Creating a prefix with pynipap results in an error being thrown from the back end if 'status' is not explicitly specified:
```
In [4]: p = pynipap.Prefix()
In [5]: p.prefix = '1.3.3.0/24'
In [6]: p.type = 'reservation'
In [7]: p.description = 'Test'
In [8]: p.save()
---------------------------------------------------------------------------
NipapError                                Traceback (most recent call last)
<ipython-input-8-395e325a0c83> in <module>()
----> 1 p.save()

/home/lukagarb/dev/venv/lib/python2.7/site-packages/pynipap.pyc in save(self, args)
   1029                     })
   1030             except xmlrpclib.Fault as xml_fault:
-> 1031                 raise _fault_to_exception(xml_fault)
   1032
   1033         # Old object, edit

NipapError: null value in column "status" violates not-null constraint
DETAIL:  Failing row contains (5625, 0, 1.3.3.0/24, 1.3.3.0/24, Test, null, null, null, reservation, 0, null, null, null, null, asdf, null, null, 0, null, {}, {}, 2014-11-07 11:04:21.945863+01, 2014-11-07 11:04:21.945863+01, 256, 0, 256, , null, infinity).
In [9]:
```

The expected behavior would IMHO be to have the prefix added with a sane default, such as 'assigned'. I'd say that the backend handles this correctly by throwing an error; pynipap should not pass the 'status' attribute to the add_prefix function when it's not been set.

plajjan:
If status hasn't been set we should default to whatever the backend
thinks ('assigned') but as pynipap used its default value of None, the
backend would try to honour that.

This changes that so we only send the status value to the backend if its
not None.

Fixes #668.

### DIFF
--- a/pynipap/pynipap.py
+++ b/pynipap/pynipap.py
@@ -981,10 +981,13 @@ class Prefix(Pynipap):
             'alarm_priority': self.alarm_priority,
             'monitor': self.monitor,
             'vlan': self.vlan,
-            'status': self.status,
             'avps': self.avps,
             'expires': self.expires
         }
+
+        if self.status is not None:
+            data['status'] = self.status
+
         for tag_name in self.tags:
             data['tags'].append(tag_name)
 


garberg:
Nice, merging!

PR #679, opened 2015-01-07
www-ui: Make sure VRF field stays populated

### LINKED ISSUE #673: VRF field not populated
Sometimes when entering the AngularJS-based add prefix page to add prefixes from a pool or other prefix, the VRF field is not populated. Annoying with intermittent faults, I have no screenshot/JavaScript error to display currently, but adding this to keep track of the issue.

garberg:
Removed a race condition which caused the VRF field on the add prefix
page to be reset when the VRF was set on page load (when adding a prefix
from another prefix or from a pool).

Fixes #673


plajjan:
Verified working. Merging.

PR #687, opened 2015-01-16
backend: fix error message for empty pool

### LINKED ISSUE #685: Assignment from empty pool gives weird error message
```
kll@lingloi320 ~ $ nipap address add family ipv6 from-pool FOO description "BAR"
Could not add prefix to NIPAP: VRF must be the same as the pools implied VRF
```

After adding a prefix to the pool it all works well. This is not the error message I would expect from an empty pool!

plajjan:
Trying to assign addresses from an empty pool would give a weird error
message ("VRF must be the same as the pools implied VRF"). Since we run
find_free_prefix after we look if the requested VRF is equal to the
pools implied VRF I have introduced a check earlier on in the code to
make sure the implied VRF is set (which means there are prefixes in the
pool). The result is an error message that is much more relatable.

Fixes #685

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -2291,6 +2291,12 @@ class Nipap:
 
             if 'from-pool' in args:
                 from_pool = self._get_pool(auth, args['from-pool'])
+
+                # make sure there are prefixes in pool, if any prefix is present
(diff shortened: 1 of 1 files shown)


garberg:
Looks good, merging! :smile:

PR #689, opened 2015-01-17
cli: use smart search for VRF listing

plajjan:
The 'nipap vrf list ...' command constructed it's own query dict instead
of relying on the smart_search_vrf method of the backend which is weird
and inconsistent as both prefix and pool listing uses the respective
smart_search functions.


garberg:
Nice nice!

PR #691, opened 2015-01-19
www: Fix node input for other allocation methods

### LINKED ISSUE #686: The node field is disabled in the "add a prefix from this pool" page when prefix type is assignment and prefix length is 32.
When I add a prefix from a pool, the node field is disabled. This is not the case in the edit prefix page.

garberg:
Fix of the incomplete fix for #675, which did not handle prefixes being
allocated from pools or other prefixes, only manual additions.

Fixes #686


plajjan:
Ok, looks good. I've tried it out as well with all the cases, i.e. manual, from-prefix and from-pool. All's well!

PR #692, opened 2015-01-19
www: require prefix field on manual assignment

### LINKED ISSUE #482: web-ui error when not defining any prefix
If you manually create a new prefix in the web-ui and don't specify the prefix this is the error message
```
invalid input syntax for type cidr: "" LINE 1: ..., NULL, NULL, '{}', NULL, NULL, NULL, NULL, NULL, '', 'nipap... ^ 
```
The web-ui should show a nice error stating that the prefix is required (marked with a *).

plajjan:
The input field has to be filled in when we are manually adding a new
prefix. Since it shouldn't be entered at all for from-pool or
from-prefix operations we use a conditional ng-required statement that
checks that the allocation method is 'manual'.

Fixes #482.


garberg:
Didn't know about that, nice solution!

Issue #693, opened 2015-01-19
Align pool statistics


plajjan:
So the pool statistics related to how many free/total prefixes are available are not consistent with how it works for the number of free/total addresses.

free/total prefixes will be NULL when the pool doesn't contain any prefixes for the address family in question or when the default prefix length for the AFI isn't set.

free/total addresses will be 0 for the above cases and not NULL.

I think the first option, i.e. being NULL, is more correct as we can more easily differentiate between a pool that is full (i.e. has member prefix for AFI and free_* is 0) and one that doesn't have a member prefix for the AFI.

Spoke IRL with @garberg who I think agree with this.

PR #694, opened 2015-01-19
cli: fix viewing of pool statistics

### LINKED ISSUE #672: cli: Tracebacks from executing "nipap pool view <pool>"
$ nipap pool view link-pni-x-x
-- Pool
  ID                         : 10
  Name                       : link-pni-x-x
  Description                : Link networks
  Default type               : assignment
  Implied VRF RT / name      : 65000:30 / xxx
  Preflen (v4/v6)            : 31 / 112
-- Extra Attributes
-- Tags
-- Statistics
  File "/usr/bin/nipap", line 122, in <module>
    cmd.exe(cmd.arg, cmd.exe_options, options)
  File "/usr/lib/python2.7/dist-packages/nipap_cli/nipap_cli.py", line 798, in view_pool
    used_percent_v6 = (float(p.used_prefixes_v6)/p.total_prefixes_v6)*100
TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'

plajjan:
To calculate correct statistics there needs to be at least one member
prefix and the default prefix length need to be set for the address
family in question. When this is not the case, the CLI would throw a
traceback but by adding some checks (similar to the ones in the web UI)
we avoid division by None and instead display a message stating that
statistics aren't available and why.

I've added another issue (#693) to track the issue that the backend
calculates statistics differently for prefixes and addresses.

Fixes #672.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -866,37 +866,65 @@ def view_pool(arg, opts, shell_opts):
 
     # statistics
     print "-- Statistics"
-    # total / used / free prefixes
-    if p.total_prefixes_v4 == 0:
-        used_percent_v4 = 0
-    else:
-        used_percent_v4 = (float(p.used_prefixes_v4)/p.total_prefixes_v4)*100
-    if p.total_prefixes_v6 == 0:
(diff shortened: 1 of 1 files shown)


garberg:
Cool, merging.

PR #697, opened 2015-01-29
Add Active Directory LDAP Auth support

### LINKED ISSUE #607: Active Directory LDAP support
The current LDAP implementation uses 'uid' ( https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap/authlib.py#L317 ), but Active Directory does not have this attribute.

dlewen:
The current LDAP Auth does not work with AD, and having the current one doing both I think would have been a bit messy, so created a new one for AD.

You use your samAccountName as username, it does UPN style LDAP bind with username@domain, where domain is from the config file.

It checks for group memberships, you have to be member of either the rogroup or rwgroup to be authenticated. If you are a member of both groups you will get rw access.

Fixes #607

### DIFF
--- a/nipap/nipap.conf.dist
+++ b/nipap/nipap.conf.dist
@@ -110,6 +110,40 @@ db_path = /etc/nipap/local_auth.db     ; path to SQLite database used
 #
 #basedn = dc=test,dc=com                ; base DN
 #uri = ldaps://ldap.test.com            ; LDAP server URI
+#
+# LDAP style
+#binddn_fmt = uid={},dc=test,dc=com
+#search = uid={}                        ; LDAP search filter
+#
+# Active Directory (UPN) style
+#binddn_fmt = {}@test.com
+#search = sAMAccountName={}
+#
+# {} in binddn_fmt is replaced by the username of the authenticating user.
+#
+## Group permissions
+# Non-empty values for rw_group/ro_group requires the memberOf attribute to be
+# present in LDAP.
+#
+# Examples:
+#
+# Everyone with an account can login and gets read/write access:
+#rw_group =
+#ro_group =
+#
+# Users in rw_group gets read/write access, everyone else gets read only access:
+#rw_group = cn=tech,dc=test,dc=com	; Users in this group get rw access
+#ro_group =
+#
+# Users in rw_group gets read/write access, users in ro_group get read
+# only access. You need to be in either group to authenticate at all:
+#rw_group = cn=tech,dc=test,dc=com      ; Users in this group get rw access
+#ro_group = cn=staff,dc=test,dc=com     ; Users in this group get ro access
+#
+# Users get read/write access by default, users in ro_group gets read
+# only access:
+#rw_group =
+#ro_group = cn=untrusted,dc=test,dc=com ; Users in this group get ro access
 
 #
 # Options for the WWW UI
--- a/nipap/nipap/authlib.py
+++ b/nipap/nipap/authlib.py
@@ -267,6 +267,10 @@ class LdapAuth(BaseAuth):
 
     _ldap_uri = None
     _ldap_basedn = None
+    _ldap_binddn_fmt = None
+    _ldap_search = None
+    _ldap_rw_group = None
+    _ldap_ro_group = None
     _ldap_conn = None
     _authenticated = None
 
@@ -294,6 +298,10 @@ class LdapAuth(BaseAuth):
         BaseAuth.__init__(self, username, password, authoritative_source, name, auth_options)
         self._ldap_uri = self._cfg.get('auth.backends.' + self.auth_backend, 'uri')
         self._ldap_basedn = self._cfg.get('auth.backends.' + self.auth_backend, 'basedn')
+        self._ldap_binddn_fmt = self._cfg.get('auth.backends.' + self.auth_backend, 'binddn_fmt')
+        self._ldap_search = self._cfg.get('auth.backends.' + self.auth_backend, 'search')
+        self._ldap_ro_group = self._cfg.get('auth.backends.' + self.auth_backend, 'ro_group')
+        self._ldap_rw_group = self._cfg.get('auth.backends.' + self.auth_backend, 'rw_group')
 
         self._logger.debug('Creating LdapAuth instance')
 
@@ -314,7 +322,7 @@ class LdapAuth(BaseAuth):
             return self._authenticated
 
         try:
-            self._ldap_conn.simple_bind_s('uid=' + self.username + ',' + self._ldap_basedn, self.password)
+            self._ldap_conn.simple_bind_s(self._ldap_binddn_fmt.format(ldap.dn.escape_dn_chars(self.username)), self.password)
         except ldap.SERVER_DOWN as exc:
             raise AuthError('Could not connect to LDAP server')
         except (ldap.INVALID_CREDENTIALS, ldap.INVALID_DN_SYNTAX,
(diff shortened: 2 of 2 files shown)


nipap/nipap/authlib.py
```diff
@@ -342,6 +343,104 @@ def authenticate(self):
...
+        except (ldap.INVALID_CREDENTIALS, ldap.INVALID_DN_SYNTAX,
+                ldap.UNWILLING_TO_PERFORM) as exc:
+            # Auth failed
+            self._logger.debug('erroneous password for user %s' % self.username)
+            self._authenticated = False
+            return self._authenticated
+
+        # Get full name and group memberships
+        try:
+            res = self._ldap_conn.search_s(self._ldap_basedn, ldap.SCOPE_SUBTREE, '(sAMAccountName=' + self.username + ')',['cn','memberOf'])
```
tobbez: The value in self.username should be escaped (by calling ldap.filter.escape_filter_chars on it) before being included in the filter string passed to search_s.

plajjan: This is a great point and applies to the already existing LDAP module as well...

tobbez:
I'm not sure whether the username needs to be escaped when performing the bind using UPN, but if it does, it should be escaped using ldap.dn.escape_dn_chars.

Aside from that and the line note, this looks good to me.

plajjan:
Ok, so we had a lengthy discussion on IRC around this PR. To sum up, I'd like to not see an extra module for AD but that we integrate the functionality needed into the current LDAP module so that it also can be used for AD authentication.

I think much of it is based on the search filter and by having something along the lines of a dn_filter variable we could use it for both scenarios;

```
AD:
dn_filter = sAMAccountName=%{username},ou=Users,dc=test,dc=com
LDAP:
dn_filter = uid=%{username},ou=Users,dc=test,dc=com
```
+ escaping et al.

The rwgroup / rogroup is new functionality (applause! :) and I would love to see the same support for "normal" LDAP, which is again a reason I don't want to have two modules. Also for the group support we should be able to handle a configuration where groups are not used, ie all users are granted RW or if we only define a rwgroup then all users get RO except for the users that are in the rwgroup who obviously gets RW access.

It also has me thinking about more granular privileges but perhaps it's not really time for that quite yet...

tobbez:
It would be possible to merge this into the LDAP module, however it's not possible to directly bind as the authenticating user in that case.

When binding against AD<sup>0</sup>, you can either use the UPN (as in this PR), or you can use the CN (usually the user's real name). This means, that to authenticate the user using their sAMAccountName, you need to:

1. Bind as another user
2. Search for the CN of the authenticating user (the filter used can be configured by the end user)
3. Unbind and try to bind using the authenticating user's CN.

This extra user would need to be configured in AD and NIPAP, and have read access to the relevant trees.

<sup>0</sup>: https://msdn.microsoft.com/en-us/library/cc223499.aspx

dlewen:
Escaping added
Merged the AD code into the LDAP module, almost as we discussed, please look at it and let me know what you think.
Groups in groups are untested, not sure how that works, if it doesn't I'll look into implementing having groups as a list in the config.

nipap/nipap/authlib.py
```diff
@@ -327,17 +335,45 @@ def authenticate(self):
...
         except:
             self.full_name = ''
+            # authentication fails if both ro_group and rw_group are configured
+            # and the search fails.
+            if self._ldap_rw_group and self._ldap_ro_group:
+                self._authenticated = False
+                return self._authenticated
+            # if either ro_group or rw_group are configured and the search
+            # fails, give readonly access
+            elif self._ldap_rw_group or self._ldap_ro_group:
```
plajjan: According to https://github.com/SpriteLink/NIPAP/pull/697/files#diff-1ac79fd5b11080d64d2820829861e54eR143 one should get RW by default, so if ro_group is defined and search fails we should give read/write, correct?

dlewen:
Maybe, depends. If one is using the ro_group its probably for some security reason, if the search fail I would prefer giving normal users read only over giving untrusted users read/write. I guess this will most likely happen if there is a LDAP miss configuration. If your normal users don't get write access they will let you know and you can fix the problem, your untrusted users will probably not tell you that they got write access by accident, so you will never know.

nipap/nipap/authlib.py
dlewen: https://github.com/dlewenMaybe, depends. If one is using the ro_group its probably for some security reason, if the search fail I would prefer giving normal users read only over giving untrusted users read/write. I guess this will most likely happen if there is a LDAP miss configuration. If your normal users don't get write access they will let you know and you can fix the problem, your untrusted users will probably not tell you that they got write access by accident, so you will never know.

tobbez: I'd agree with giving read only rather than read/write access here.

However, this except clause will only be reached if search_s fails, which should never happen, except when it's passed an invalid filter string (_ldap_search is misconfigured), or when there's a problem communicating with the LDAP server (e.g. connection has dropped since binding). The former is likely to be much more common.

Maybe it would be better to raise an exception instead?

nipap/nipap/authlib.py
```diff
@@ -327,17 +335,45 @@ def authenticate(self):
 
         # auth succeeded
         self.authenticated_as = self.username
-        self._authenticated = True
         self.trusted = False
         self.readonly = False
 
         try:
-            res = self._ldap_conn.search_s(self._ldap_basedn, ldap.SCOPE_SUBTREE, 'uid=' + self.username, ['cn'])
+            res = self._ldap_conn.search_s(self._ldap_basedn, ldap.SCOPE_SUBTREE, self._ldap_search.format(ldap.dn.escape_dn_chars(self.username)), ['cn','memberOf'])
```
tobbez: This search should not be performed if both ro_group and rw_group are empty. That would make it possible to use empty ro_group/rw_group to provide everyone with read/write access if you don't have the memberOf attribute in your LDAP instance.

Currently, the search_s call would raise an NO_SUCH_ATTRIBUTE exception, meaning all authenticated users would get read only access instead.

nipap/nipap/authlib.py
plajjan: I'm sorry, I read this on a mobile device and didn't get the whole context when I initially commented.

If we only end up in here because of something like an invalid filter string, we should throw an exception instead and please try to refrain from using just except - try to match on the specific exception instead.

nipap/nipap.conf.dist
```diff
@@ -110,6 +110,40 @@ db_path = /etc/nipap/local_auth.db     ; path to SQLite database used
...
+# Users in rw_group gets read/write access, everyone else gets read only access:
+#rw_group = cn=tech,dc=test,dc=com
+#ro_group =
+#
+# Users in rw_group gets read/write access, users in ro_group get read
+# only access. You need to be in either group to authenticate at all:
+#rw_group = cn=tech,dc=test,dc=com      ; Users in this group get rw access
+#ro_group = cn=staff,dc=test,dc=com     ; Users in this group get ro acces
+#
+# Users get read/write acces by default, users in ro_group in gets read
```
plajjan: acces = > access
"users in ro_group in gets" << remove extra "in"

nipap/nipap.conf.dist
```diff
@@ -110,6 +110,40 @@ db_path = /etc/nipap/local_auth.db     ; path to SQLite database used
...
+#ro_group =
+#
+# Users in rw_group gets read/write access, everyone else gets read only access:
+#rw_group = cn=tech,dc=test,dc=com
+#ro_group =
+#
+# Users in rw_group gets read/write access, users in ro_group get read
+# only access. You need to be in either group to authenticate at all:
+#rw_group = cn=tech,dc=test,dc=com      ; Users in this group get rw access
+#ro_group = cn=staff,dc=test,dc=com     ; Users in this group get ro acces
```
plajjan: acces => access

nipap/nipap/authlib.py
dlewen: Not sure I follow, that's exactly what is intended, if both ro_group and rw_group is empty, no group member security is used, and everyone will always get read/write access no matter if memberOf is present or not. 

When I was testing I did not get NO_SUCH_ATTRIBUTE from the search_s if the memberOf attribute was missing, however I do get KeyError on 'if self._ldap_ro_group in res[0][1]['memberOf']:'

But sure, we can do a separate search for memberOf only if at least one of ro_group or rw_group is non empty.

nipap/nipap/authlib.py
dlewen: Sure, do you happen to know what ldap exceptions I can get from a search_s? Or should I use LDAPError? We can also get KeyError if the memberOf attribute is missing at `if self._ldap_ro_group in res[0][1]['memberOf']:`. At least from my tests the search_s succeeds when the attribute is missing.

nipap/nipap/authlib.py
tobbez: Never mind then. From a cursory look at the python-ldap it seemed like it would raise the NO_SUCH_ATTRIBUTE exception when you request an attribute that doesn't exist, but since it doesn't, it doesn't matter. I also overlooked that the search result was used to get the name of the user, so it seemed as if the result wasn't used at all if groups weren't non-empty. 

Just disregard my previous comment - doing just one search is fine.

Adding a specific except clause for the KeyError, that raises an appropriate exception would probably be a good idea though (since it indicates misconfiguration).

plajjan:
Dunno if you are interested in the python3 compatibility effort going on over in #611. I don't mind keeping the two issues separate and convert the LDAP implementation to ldap3 once this PR is merged, but if you are interested in writing for py3 I would happily merge a PR that fixes both AD and py3 compatibility in one go.

Regardless, before merging anything I want a squashed history - no need to see the rejected code in our history.

nipap/nipap/authlib.py
dlewen: We will also end up in the except if the username does not exist at all (but then the bind should fail) or if the basedn is wrong, or if the username is not found with the configured basedn.

dlewen:
I prefer keeping it separate, it will be to much work for me right now to test this code with py3.

PR #698, opened 2015-01-29
Don't use mutable default arguments

tobbez:
It's generally a bad idea since they are only instantiated once (at declaration time).

An example:
```python
>>> def tst(arg=[]):
...     arg.append(len(arg))
...     print arg
...
>>> tst()
[0]
>>> tst()
[0, 1]
>>> tst()
[0, 1, 2]
```

### DIFF
--- a/nipap/nipap/authlib.py
+++ b/nipap/nipap/authlib.py
@@ -124,7 +124,7 @@ class AuthFactory:
 
 
 
-    def get_auth(self, username, password, authoritative_source, auth_options={}):
+    def get_auth(self, username, password, authoritative_source, auth_options=None):
         """ Returns an authentication object.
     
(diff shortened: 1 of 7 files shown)


plajjan:
So, there's tests/db_free.py that has a tags=[] value that is not touched by this PR. db_free.py is not run by Travis-CI, so not really used at this point but perhaps we should just fix everything.

Would you like to fix that (please squash the commits) and I'll happily merge this.

tobbez:
I would, but I can't seem to find the file you're referring to (or any with a similar name). I found a couple of others I missed and fixed those though. There shouldn't be any instances left now.

plajjan:
Heh, db_free.py is a local file. I did some work on that in the past and forgot to check in to git.

Anyway, great job. It's a good catch. Merging.

PR #699, opened 2015-02-02
[requires.io] dependency update on master branch

plajjan:
This is an automatically generated PR from requires.io.


plajjan:
I think this fails due to psutil 2.x being installed while we rely on 1.x. Syntax is completely different.

We use 1.x because that is what debian is shipped with.

plajjan:
We could probably consider the other requirements though.

plajjan:
I'll look into writing some filters for the requirements.txt file which should stop requires.io to open PRs with broken deps.

PR #719, opened 2015-03-26
whoisd: add IPv6 support

### LINKED ISSUE #718: whoisd crashes when listening on IPv6
When the config file contains `listen = ::` (or `listen = [::]`), nipap-whoisd crashes at startup:

```
$ nipap-whoisd --config /etc/nipap/whoisd.conf
  File "/usr/sbin/nipap-whoisd", line 180, in <module>
    except socket.error, exc:
NameError: name 'socket' is not defined
```

If I change the config file to say `listen = 0.0.0.0`, it starts fine. The full whoisd.conf is found below:

```
[whoisd]
user = nobody
group = nogroup
listen = ::
port = 43
pid_file = /var/run/nipap-whoisd/whoisd.pid
nipapd_host = localhost
nipapd_port = 1337
nipapd_username = whoisd-user
nipapd_password = foo
```

plajjan:
SocketServer is hard-coded to listen to IPv4 so this introduces some
monkey patching to get around that and enable it for listening on an
IPv6 socket. When binding to "" or :: the server will listen on both
IPv4 and IPv6. To listen on an IPv4 address it needs to be specified as
an IPv4-mapped IPv6 address.

Fixes #718.

### DIFF
--- a/whoisd/nipap-whoisd
+++ b/whoisd/nipap-whoisd
@@ -8,6 +8,7 @@ import sys
 
 import ConfigParser
 import SocketServer
+from SocketServer import ForkingMixIn, TCPServer
 
 import nipap_whoisd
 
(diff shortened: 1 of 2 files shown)


garberg:
Nice! IPv6 \o/

PR #725, opened 2015-04-03
Upgrade AngularJS to v1.3.15

### LINKED ISSUE #680: Upgrade AngularJS to 1.3.8
We are currently using AngularJS 1.2 and there are a number of new features and fixes in AngularJS 1.3 that we probably want.

From the release notes of 1.3:
```Angular 1.3 provides a boost in performance versus the previous versions. The Angular development team concentrated on making it less overhead. According to Jeff Cross and Brian Ford at the Ng-Europe conference, they achieved 4.3 times faster DOM manipulation with 73% less garbage and 3.5 times faster digest with 87% less garbage versus 1.2.0 version based on their largetable benchmark tests.```

... this sounds pretty nice. The performance (or lack thereof) is quite noticeable on the prefix list page.

Another sweet feature is the possibility of doing one-way bindings, thereby improving performance considerably for use cases where two-way bindings is not strictly necessary. Most of our listing pages, most notably the prefix list page, should be able to gain from this.

I'm not sure at all how we should test 1.3. Do we need to at all?

garberg:
Upgraded AngularJS (core and ngRoute) to version 1.3.15.

Fixes #680


plajjan:
As always with Angular related stuff, I have a real hard time reviewing PRs but overall it looks good to me.

We still have the behaviour that the VRF selector dialogue closes when clicked. Should we fix that too or do you want to keep that in a separate issue (do we have an issue for that?)?

plajjan:
I added #728 to keep track of the VRF selector closing issue. Since the rest of this patch looks good I'm merging! =)

PR #726, opened 2015-04-04
More additional LDAP support

CrackerJackMack:
Adds the cacert and tls options to LdapAuth.

cacert should be the full path to the cacert used on the LDAP server
tls should be a truthful value (non-zero) to force TLS

Also fixed a bug with the global name ldap not being available

### DIFF
--- a/nipap/nipap.conf.dist
+++ b/nipap/nipap.conf.dist
@@ -110,6 +110,7 @@ db_path = /etc/nipap/local_auth.db     ; path to SQLite database used
 #
 #basedn = dc=test,dc=com                ; base DN
 #uri = ldaps://ldap.test.com            ; LDAP server URI
+#tls = False                            ; initiate TLS, use ldap://
 #
 # LDAP style
 #binddn_fmt = uid={},dc=test,dc=com
--- a/nipap/nipap/authlib.py
+++ b/nipap/nipap/authlib.py
@@ -75,6 +75,12 @@ import sqlite3
 import string
 import random
 
+try:
+    import ldap
+except ImportError:
+    ldap = None
+
+
 class AuthFactory:
     """ An factory for authentication backends.
     """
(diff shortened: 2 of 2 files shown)


plajjan:
Isn't this configurable system-wide via /etc/ldap/ldap.conf?
Does this need to be configurable for NIPAP, isn't it just enough to do it per system? One typically adds their enterprise CA cert there anyway, no? :)

CrackerJackMack:
I have it setup in /etc/ldap.conf and /etc/ldap/ldap.conf to no avail.  I know these are working because pam uses this for authentication.  Using ldaps:// does not work without setting the OPT_X_TLS_CACERTFILE option.  same with enabling TLS doesn't work with that cacert option and callig start_tls_s()

Do you have a working system with TLS and a private CA using only ldap.conf ?  If so, I'm at a loss as to why these don't work.

Another reason to use these is for hosting multiple nipap instances for different user groups.  This might be useful during a merger and acquisition until the spaces/processes can be merged as well.

garberg:
At the main NIPAP installation I administer we use LDAP authentication over SSL(```uri = ldaps://...``` in nipapd.conf). What's in your ldap.conf? Mine:
```
TLS_CACERT     /etc/ldap/ca.crt
```
I'd be happy to merge the LDAP-fixes, but if the SSL-stuff can be fixed system-wide it'd be great to avoid adding it to NIPAP.

What version of the python-ldap package do you have?

CrackerJackMack:
Wow, so we have `tls_cacertfile` in our `/etc/ldap.conf` and `/etc/ldap/ldap.conf`.  Adding `tls_cacert` to `/etc/ldap/ldap.conf` made nipap work.  Using `python-ldap==2.4.19` on ubuntu 12.04.

Which is odd, because everything else is working.  I guess it just goes to show how ldap is a bit wonky still.  Have any hints for using TLS instead of ldaps ?  Or should I re-base this commit w/o the cacert settings?

plajjan:
I forgot to mention before, thank you for your effort and nice code! :)

I suppose that the scenario you describe with multiple nipap instances would be towards multiple different LDAP servers with potentially different self-signed certs and you would therefore need multiple CAs, correct?
I think you can just put all of those certs into one file and point to that with tls_cacert, or use the tls_cacertdir to point to a directory and keep one cert per file.

For the TLS part, I think having an option is the only sensible way. We could try to always establish TLS but if the server doesn't support it (or CA isn't trusted), then what would we do? If we silently accept non-TLS connection, the administrator might not realize he is running a non-secure setup. If we always require TLS then we would have to drop the connection. I suppose that running unencrypted is quite common for a lab setup or just during deployment to make sure the basic stuff is working. I see no other way but to have an option for it. I think the default should be to require TLS.

And yes, if we can achieve the same thing around cacert through ldap.conf then I don't think we should add extra arguments to NIPAP just for the sake of adding, so please rebase/modify without cacert setting.

CrackerJackMack:
squashed those commits as requested

plajjan:
I've done some limited testing which shows no problems. Code looks good and we know it fixes at least one bug. Merging.

PR #729, opened 2015-04-06
Make pool and vrf names case-insensitive unique

### LINKED ISSUE #341: Add case-insensitive uniqueness to VRF name
Should the VRF name be uniqueness be case-insensitive?

plajjan:
Not really sure how we should handle upgrades though!? Perhaps just
write something in the README that people need to think about this.

Fixes #341.


garberg:
Looks nice, merging!

PR #737, opened 2015-04-12
Update config examples for LDAP

plajjan:
Replace 'test.com' with 'example.com'.

The examples for how to configure LDAP, in particular the basedn was not
particularly clear since it didn't include ou=Users, which is typically
used.

### DIFF
--- a/nipap/nipap.conf.dist
+++ b/nipap/nipap.conf.dist
@@ -108,23 +108,29 @@ db_path = /etc/nipap/local_auth.db     ; path to SQLite database used
 #[auth.backends.ldap1]
 #type = LdapAuth
 #
-#basedn = dc=test,dc=com                ; base DN
-#uri = ldaps://ldap.test.com            ; LDAP server URI
+#basedn = ou=Users,dc=example,dc=com       ; base DN
+#uri = ldaps://ldap.example.com            ; LDAP server URI
 #tls = False                            ; initiate TLS, use ldap://
 #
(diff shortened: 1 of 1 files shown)


plajjan:
Looking for input from @alexrasmussendk and @CrackerJackMack on this one.

CrackerJackMack:
lgtm   reference: http://tools.ietf.org/html/rfc2606#section-3

CrackerJackMack:
based on the conversation in IRC today we should probably mention that in some cases the LDAP servers are case sensitive and sometimes the non-OU basedn (dc=example,dc=com) search's aren't always allowed.

plajjan:
Indeed, I hadn't understood the case-sensitive part when I wrote the patch. While I think the updated examples (including ou=Users) should remove some cause of mistakes it might be a good idea to point out why.. are we sure this is due to recursion not being supported or could it be something else?

CrackerJackMack:
Since I'm not able to reproduce this myself but we do have the corresponding error and solution we can reference them both in the config example.

plajjan:
The message is 
```
Authentication error: {'info': '000004DC: LdapErr: DSID-0C0906E8, comment: In order to perform this operation a successful bind must be completed on the connection., data 0, v1db1', 'desc': 'Operations error'}
```

As discussed on IRC, I think we'll add it to a FAQ as to not overload the configuration file too much. I'm keeping it here for reference.

PR #739, opened 2015-04-13
Make "status" action of init scripts LSB compliant

toreanderson:
Previously, the init scripts would unconditionally return 0 when called
with the "status" action, which misleads (non-human) callers into
believing that the service in question is running even though it is not.
Fix that by explicitly exiting the init script with the appropriate
return code. Also change the slightly improper usage of log_action_msg
with log_success_msg/log_failure_msg.


plajjan:
Looks good to me - Nice!

PR #741, opened 2015-04-13
Add example infoblox import script

CrackerJackMack:
Adds example infoblox utility that isn't 100% perfect but does the needful and adds all the attributes as tags.  The avps property could be explored as well but this is just an example utility.


plajjan:
Thank you very much!

Not much I can test since I don't have a InfoBlox system. The only feedback I have would be to make credentials for Infoblox & NIPAP configurable, but it's an import script.. likely only run once or a few times. People can figure it out. Merging.

PR #742, opened 2015-04-15
pass port configuration paramater to driver

giganteous:
Obviously no one else uses a different port :-)


giganteous:
I didn't write any tests. I didn't look for them. Should I take the time for that?

plajjan:
Heh, yeah, obviously everyone is running default port. No need for tests on this one! Thank you :)

PR #748, opened 2015-04-22
Fix confirmation message for when editing prefixes

toreanderson:
Also fixes an incorrectly spelled word in a comment nearby.


plajjan:
Thanks!

PR #752, opened 2015-05-02
www: fix expires field of infinity

### LINKED ISSUE #721: expire usage bug (invalid defaults)
when testing nipap for a massive import/conversion of our current IPAM I ran into the following issue using nipap-www.  When the expire field is missing it's default is `None` is is then represented as `str(self.expire)`.  This makes the default value in the webform invalid. [related commit](https://github.com/SpriteLink/NIPAP/pull/658/files#diff-26330bff8775394301a1bdaeef7e5d43R860)

In a related matter clearing the expires field sets the expire date to midnight.  It appears that setting the `expires = None` using pynipap doesn't quite have the desired results, which is always setting the expired attribute even when you don't want it to expire.

plajjan:
There were two issues surrounding the handling of positive infinite
values for the expires field. Upon reading the values the JSON encoder
would always cast to string which doesn't work very well with None
values as that would be represented as the string "None" rather that
<nil> over the wire and as we use nil for presenting infinity, this
would mess things up. Properly handling this means the text input in the
web form stays empty.

Secondly, we mangle the expires field to format the field into a ISO8601
value before sending to the backend when editing/adding a prefix. We
shouldn't mangle if the expires field is empty, since this means
infinity and mangling it will mess it up.

Fixes #721.

### DIFF
--- a/nipap-www/nipapwww/controllers/xhr.py
+++ b/nipap-www/nipapwww/controllers/xhr.py
@@ -879,6 +879,9 @@ class NipapJSONEncoder(json.JSONEncoder):
 
             vrf_id = obj.vrf.id
             vrf_rt = obj.vrf.rt
+            expires = None
+            if obj.expires is not None:
+                expires = str(obj.expires)
 
             return {
                 'id': obj.id,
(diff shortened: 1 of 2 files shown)


garberg:
Looks nice, merging!

PR #753, opened 2015-05-02
Fix updating of the last_modified value

plajjan:
The last_modified value was in fact not updated when modifying a prefix, in most cases at least. This was due to conditions on the trigger that performs the update making it run only when the vrf_id, type or prefix was changed. The conditions themselves are there to increase performance by avoiding to run a large trigger function when it's not necessary.

By moving the conditional part from the trigger definition into the trigger function, the last_modified value can always be updated but the lions share of the function avoided when it is not needed.


garberg:
:thumbsup:

PR #574, opened 2014-08-12
Remove unused code

dosomder:
Remove unused code according to scrutinizer-ci

PR #756, opened 2015-05-05
Modifications of dosomder's "Remove unused code"

garberg:
As we wanted some changes to @dosomder's pull request #574 and he stated that he unfortunately cannot spend time on the project at the moment, I took the liberty to create a new pull request including the original commit and some further changes.

### DIFF
--- a/nipap-cli/nipap_cli/command.py
+++ b/nipap-cli/nipap_cli/command.py
@@ -22,9 +22,9 @@ class Command:
 
 
     key = {}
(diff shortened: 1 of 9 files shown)


plajjan:
Not much to say about this, it's pretty straight forward. We did have a discussion at some point how we should do imports from pynipap, like should we always import VRF, Pool and Prefix or only import what we need? I don't think we ever reached consensus on that. 

Merging.

Issue #750, opened 2015-04-27
API doc incorrectly says edit_prefix takes "spec" argument


toreanderson:
From http://nipap-plajjan.readthedocs.org/en/latest/nipap.html#nipap.backend.Nipap.edit_prefix:
```
spec [prefix_spec]
    Specifies the prefix to edit.
```
That doesn't work, however. This should probably read `prefix [prefix_spec]` instead (which does work).

For what it's worth I see that the `spec` attribute is mentioned for other API calls too, so maybe the same mistake is made elsewhere.

plajjan:
Ah yes. Technically, the docs are correct. You are looking at the docs for the backend module and its argument is called spec. However, we have the XML-RPC wrapper around the backend to expose its methods over the XML-RPC API and that method  takes an argument called 'prefix'. I believe if you look at that methods docs you will find that the argument is called prefix. 

Since we want to avoid duplication we have tried to put as much documentation in the backend module so that readers should not have to look around in a lot of places to understand the design. This is an example where the reader would have to look around in multiple places and there is no reason for it. I suggest we simply rename the argument in one of the functions so that they are called the same.

toreanderson:
Ah, quite right you are! I hadn't realised those were two distinct parts of the documentation, sorry about the noise. If you want to close this issue, do feel free.

FWIW the way I found out that `prefix` was used instead of `spec` was by resorting to tcpdumping the XML-RPC traffic between the web interface and the backend. :-)

plajjan:
No, I won't close this. I think it's excellent proof that the current documentation is not clear. In addition, you are referencing my private build of the docs instead of the official docs, which made me notice that the API docs is actually missing in the official docs (as per #754).

Not quite sure how we should solve it though.
* Better cross referencing between the "NIPAP API" (which is really just the internal backend module) and the XML-RPC API (which most people will expect)
* Have complete docs for the XML-RPC API. The reason I don't like this is because if we do add a second wrapper around the backend module, like a JSON-RPC API, we would have to completely duplicate all the documentation on how a particular module behaves.
* Something else?

garberg:
I have done some work on pynipap similar to this while looking into #712. It includes your first point, cross references from pynipap functions to their respective backend function. It works fine building the docs locally using sphinx, but when building on readthedocs the references are simply not turned into links from some reason.

Oh, now I see that the NIPAP API part is missing entirely as described in #754, that might be related :laughing: 

[...]

PR #758, opened 2015-05-06
Add references to backend functions in pynipap docs

garberg:
To avoid duplicating information the pynipap documentation is very brief as most data is passed straight to the backend anyway. This PR tries to make this more clear by adding references to the corresponding backend function to each of the pynipap class methods which involve backend operations.


garberg:
The result can be seen at http://nipap-garberg.readthedocs.org/en/712-update_pynipap_docs/.

garberg:
@plajjan, any comments to this?

plajjan:
Well, yes, I just have difficulty forming an opinion :P 
Overall I think having references is a good thing but given issues like #750 I'm not convinced we should be linking to the "NIPAP API", which also happens to be a name I don't like, I don't think we should call it an API even if one could consider as such. It is not meant to be interfaced with by normal users. Renaming it isn't necessarily part of this PR though.

So, to focus on your PR, I think it needs to be made clear that the XML-RPC layer is in between pynipap and the NIPAP backend module. Either by linking first to XML-RPC from pynipap or by expressing it in NIPAP API and from NIPAP API linking to XML-RPC API for every function.

plajjan:
Just the wording "For more information, see the backend funtion X" is not a good one. I know you added a piece at the top of the documentation about how we want to keep things in pynipap very "thin" but chances are people are not going to read the documentation top-down. They will go look at the very function they are interested in and for that reason I think the wording of the text that refers to the backend module needs to be understandable on its own. I base this on how I typically read docs ;)

Something more along the lines of:
"pynipap maps directly to the XML-RPC function X, please see the documentation of that function for information on input arguments and return values"

Or "pynipap maps to function X in the backend, please see the documentation of that function for information on input arguments and return values"

In addition, in the backend module, perhaps we should have a section explaining the XML-RPC API and how it wraps around the backend and for every backend function we have a reference saying something along the lines of "This is the documentation of the internal backend function. It is exposed over XML-RPC, please also see the documentation of the XML-RPC function for full understanding.".

PR #761, opened 2015-05-07
www: allow deselection of selected pool

### LINKED ISSUE #732: www-ui: Not possible to unselect an already selected pool when adding a new prefix
Not possible to unselect an already selected pool when adding a new prefix.

Is this information enough? :)

plajjan:
When adding a prefix, the pool it is a member of can be specified,
however the page didn't allow you to deselect the pool once it was
selected. That is now fixed.

Fixes #732.


garberg:
:thumbsup:

PR #763, opened 2015-05-07
www: fix incorrect close of vrf and pool selector

### LINKED ISSUE #728: VRF selector closes when clicked
The VRF selector widget closes when it is clicked. The only reason it actually works right now is because it receives focus once the "Change VRF"  button is clicked and the VRF Selector widget is opened.

plajjan:
The VRF and pool selector would close when clicked on which isn't very
intuitive at all. Found a stackoverflow discussion on this topic:
http://stackoverflow.com/questions/23698316/how-to-prevent-an-angular-bootstrap-dropdown-from-closing-unbind-event-which-wa

This implements that fix for both VRF and pool selector.

Fixes #728.


garberg:
Oh, neat solution! :smile:

PR #770, opened 2015-05-18
cli: accept natural mask for prefix view

### LINKED ISSUE #738: Make view-command accept display prefix
Today it's pretty annoying that host prefixes shown as result from the ```nipap address list ...``` command can not be copied and directly used with the ```nipap address view ...``` command as the former shows the display prefix attribute and the latter matches the prefix against the prefix attribute.  

One solution would be to let the view-command match its input against ```prefix or display_prefix```. Would that have any unwanted side-effects?

plajjan:
Previously the view prefix function wouldn't accept the display_prefix,
i.e. a prefix with its natural mask. To show a host prefix it had to be
entered either with a host mask, e.g. /32 or /128 or with no mask at all.

There are scenarios when a host has the network address, most typically
when using /31 or /127 networks. To distinguish between the two, the
mask must be entered, i.e. use /31 to show the network or /32 to show
the host.

Fixes #738.

I did contemplate putting in a warning and/or hint saying that users
need to be careful when using a network address since it could
potentially hide a host prefix but decided not too as ultimately users
should know what they are doing.  I think we should add a section to the
NIPAP user guide CLI section about this thoug.

This also introduces a requirement on IPy for the CLI, which we haven't
had before but I don't think that's a big deal. I've been contemplating
including IPy before anyway, like we could return Pynipap.Prefix.prefix
as a IPy object instead of a text string..

Test runs through despite the lack of IPy in setup.py because IPy is
already installed in Travis-CI as it's required by the backend.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -981,7 +981,7 @@ def view_prefix(arg, opts, shell_opts):
     try:
         # this fails if bits are set on right side of mask
         ip = IPy.IP(arg)
-    except:
+    except ValueError:
         arg = arg.split('/')[0]
 
     q = { 'prefix': arg }
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -16,6 +16,8 @@ import string
 import subprocess
 import sys
 
+import IPy
+
 import pynipap
 from pynipap import Pool, Prefix, Tag, VRF, NipapError
 from command import Command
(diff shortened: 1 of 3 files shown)


plajjan:
Hmm, I think I've changed my mind, I will include dependency update in this PR. Hold on a sec :)

nipap-cli/nipap_cli/nipap_cli.py
```diff
@@ -966,6 +968,21 @@ def view_pool(arg, opts, shell_opts):
...
+    # to assume /32 if the address is not a network address. If it is the
+    # network address we always search using the specified mask. In certain
+    # cases there is a host with the network address, typically when /31 or /127
+    # prefix lengths are used, for example 1.0.0.0/31 and 1.0.0.0/32 (first host
+    # in /31 network) in which case it becomes necessary to distinguish between
+    # the two using the mask.
+    try:
+        # this fails if bits are set on right side of mask
+        ip = IPy.IP(arg)
+    except:
```
garberg: Hm, it's always a bit dangerous to catch all types of exceptions. I'd prefer catching a ValueError (which my limited empirical studies of the IPy module suggests is the exception thrown for invalid prefixes) explicitly!

garberg:
NICE!

PR #772, opened 2015-05-18
nipapd: setup.py soft-fail when lacking rst2man

plajjan:
Use a special build class to dynamically build the list of data_files,
this allows us to try and generate man files but in the absence of
rst2man, the man files will simply not be included.

rst2man (docutils) is required to build but not install or run NIPAP so
I don't want to list docutils as a dependency in setup.py. We are
currently using distutils but if we were to switch to setuptools then we
would have the option of specifying setup_requires and install_requires
separately. I did try to rewrite this for setuptools but couldn't get it
to work without additional changes so I gave up on that.

I'm not sure if we should do this or not.. it's kind of a philosophical
question. Should we fail-hard when rst2man is not present? It would
allow to build incomplete .deb files. On the other hand I usually build
the deb files and I have rst2man installed so it's not much of an issue
and for anyone that just wishes to install NIPAP, not build packages, it
could be convenient that it's possible to do so from source without
first manually installing rst2man.

@garberg, what do you think?


jessesanford:
can we get this merged? Looks pretty good to me

PR #777, opened 2015-05-18
Fix pool tag searches

plajjan:
This fixes pool smart searching for tags.


garberg:
Looks great, merging!

PR #778, opened 2015-05-19
Add scheme to the CLI for HTTPS URI's

CrackerJackMack:
The desire was to add https support but since xmlrpc already
supports it, we just increased the options section to
include 'scheme' and added the default to 'http' since we
don't want to break older clients.  Hopefully this change
is transparent and non-breaking.

### DIFF
--- a/nipap-cli/nipap
+++ b/nipap-cli/nipap
@@ -74,7 +74,11 @@ if __name__ == '__main__':
         sys.exit(1)
 
     # read configuration
-    cfg = ConfigParser.ConfigParser()
+    cfg_defaults = {
+        'scheme': 'http',
+        'port': '1337',
+    }
+    cfg = ConfigParser.ConfigParser(cfg_defaults)
     cfg.read(userrcfile)
     nipap_cli.nipap_cli.cfg = cfg
 
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -48,11 +48,12 @@ def setup_connection():
 
     # build XML-RPC URI
     try:
-        pynipap.xmlrpc_uri = "http://%(username)s:%(password)s@%(hostname)s:%(port)s" % {
+        pynipap.xmlrpc_uri = "%(scheme)s://%(username)s:%(password)s@%(hostname)s:%(port)s" % {
+                'scheme': cfg.get('global', 'scheme'),
                 'username': cfg.get('global', 'username'),
                 'password': cfg.get('global', 'password'),
                 'hostname': cfg.get('global', 'hostname'),
-                'port'    : cfg.get('global', 'port')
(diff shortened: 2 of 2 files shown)


nipap-cli/nipap
```diff
@@ -74,7 +74,12 @@ if __name__ == '__main__':
         sys.exit(1)
 
     # read configuration
-    cfg = ConfigParser.ConfigParser()
+    cfg_defaults = {
+        'scheme': 'http',
+        'hostname': 'localhost',
```
plajjan: Do we really need a default value for hostname? Users have to define username and password. Is it likely that NIPAP will be running on the same host as the CLI is running on?

nipap-cli/nipap
```diff
@@ -74,7 +74,12 @@ if __name__ == '__main__':
...
+        'port': '5000',
```
plajjan: Why 5000? nipapd binds to 1337 per default. It's paster for the web-ui that does 5000 per default, but nipap-cli doesn't talk to the web ui..

CrackerJackMack: That is an oversight.  I will fix that.

nipap-cli/nipap
```diff
@@ -74,7 +74,12 @@ if __name__ == '__main__':
         sys.exit(1)
 
     # read configuration
-    cfg = ConfigParser.ConfigParser()
+    cfg_defaults = {
+        'scheme': 'http',
```
plajjan: Is this a "scheme"? I'm trying to think what I would intuitively call it... Isn't like SSL: on/off or True/False better and more intuitive?

plajjan:
I agree with the overall goals, we obviously don't want to run over unencrypted transport, but I'm not convinced by the code as it stands.

Also, and now I'm nitpicking, the subject of the commit should probably be a little more specific :) We could fit a lot of the code committed under that subject ;)

nipap-cli/nipap
CrackerJackMack: For those trying out nipap in development or vagrant this is handy, but not necessary.  I can remove it if that's the decision and just leave the port and scheme in the defaults since those should rarely be specified.

nipap-cli/nipap
CrackerJackMack: the word scheme wasn't pulled out of thin air: http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml  It's part of a URI spec.

Since the current config is based off the URI parts and not a class constructor I say we stick to the international standards for now until we wrap up the initialization into something else.

CrackerJackMack:
Agreed that the commit is too broad, I'll adjust that as well.

garberg:
Not sure why this PR was left hanging, if you rebase on a recent master I'd be happy to merge.

garberg:
Feature added in #1283

PR #781, opened 2015-05-20
backend: update allowed attributes for prefix spec

### LINKED ISSUE #667: Update allowed attributes in _expand_prefix_spec
The backend function _expand_prefix_spec performs a validation of what prefix attributes are allowed by comparing received attributes with a list. A few attributes, for example order_id, are missing. Make sure that the list is up to date!

plajjan:
Add a global list of all attributes of a prefix, also marking which ones
are read-only or read-write, which we derive a second list from that
only contains the writable attributes.

These two lists are then used in all the places where we check input
arguments.

Fixes #667.

@garberg, what do you think of this approach? I didn't apply the same
for pools and VRFs because I wanted input on this one first but if you
think this looks good we should do the same for them as well.

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -209,6 +209,133 @@ except:
     import parsedatetime.parsedatetime_consts as pdc
     pdt = parsedatetime.parsedatetime.Calendar(pdc.Constants())
 
+
+# list of all attributes on a prefix, including both writable and read-only
+# values
+_prefix_spec = {
+        'added': {
+            'column': 'inp.added',
+            'ro': True,
+        },
+        'alarm_priority': {
+            'column': 'inp.alarm_priority',
+            'ro': False,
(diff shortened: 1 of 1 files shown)


garberg:
Generally it looks great, a really nice improvement!

But, as the PR is currently implemented, isn't the ability to search for read-only attributes destroyed, as the ```_prefix_attrs``` filters out them? Or am I missing something? Perhaps there should be a ```_prefix_attrs_ro``` with the read-only attributes and a ```_prefix_attrs_rw``` with the writable attributes, which can be concatenated when performing validations in the ```search_```- and ```list_```-functions?

plajjan:
No, because the search functions look in _prefix_spec while add_prefix and edit_prefix look in _prefix_attr, which is a subset of _prefix_spec.

garberg:
Doh.

PR #782, opened 2015-05-20
backend: enable case-insensitive tag searches

### LINKED ISSUE #743: Case-insensitive tag search
The tag search is today case-sensitive. Make it case-insensitive!

plajjan:
PostgreSQL doesn't have very powerful array operators so doing a
case-insensitive search is actually kind of difficult by normal means
(like lower()). Instead I use the citext module which is a standard
extension, shipped with PostgreSQL since 8.x, that allows
case-insensitive matches. This in turn makes it easy to recast the tags
column as citext and match on.

Fixes #743.

@garberg, the alternative approach is that we simply turn the tags
column into a citext array instead of the text array that it is today.
That would however remove the option of doing case-sensitive searches on
the tags column.. well, we could recast it the other way, ie from citext
to text[] if we actually wanted a case-insensitive match.

There are no indices on the tags column today.. perhaps we should add that
too.


garberg:
We discussed this quite extensively offline, pondering whether all text columns should be citext, if all our searches by default should be case insensitive and so on. I think the conclusion we finally came to was we'll merge this as it is and then later change all search operators to be case insensitive, as is the case with the reqular expression operator. I also think we came to the conclusion that we'd not change the datatypes in the database from text to citext. Performace-wise it'd not matter as we'd create a functional index with the data casted to citext on the columns anyway.

Issue #683, opened 2015-01-11
Improve query interpretation format


plajjan:
The current interpretation format is not flexible and powerful enough to describe more complex queries. It makes certain assumptions but with the introduction of a more complex query language, those assumptions will fail.

This is closely related to #531 but I'm splitting it out as this can be implemented on it's own and I think it is easier for the everyone with as small changes as possible.

The interpretation format needs to be changed as it is not flexible enough to express more complex queries. I would suggest that the interpretation should be baked in together with the dictSQL struct - the two structures are so similar so having them separate is just a waste. In the combined format there would be certain keys that pertain to the core functionality of dictSQL (ie forming a query that can be translated into SQL) and we would have other keys that help with printing an interpretation.

For example, to match within a prefix we have the following dictSQL:
```
{
   'operator': 'contained_within_equals',
   'val1': 'prefix',
   'val2': '1.0.0.0/24'
}
```

Including the interpretation would make it look like:
```
{
   'operator': 'contained_within_equals',
   'val1': 'prefix',
   'val2': '1.0.0.0/24',
   'interpretation': {
      'interpretation': 'IPv4 prefix',
      'string': '1.0.0.3/24',
      'attribute': 'prefix',
      'text': 'Prefix must be contained within 1.0.0.0/24, which is the base prefix of 1.0.0.3/24.',
      'operator': 'contained_within_equals',
      'strict_prefix': '1.0.0.0/24'
   }
}
```

All the keys are copied from the current interpretation format, except for "text" which I've added. In our current implementation we let the client (web or CLI) parse the interpretation dict and format a suitable text message but I'm leaning towards doing this on the server side to avoid writing as much code in both clients. The reason for the current implementation is that a) at first it only existed in the web UI and b) I wanted to be able to put stuff in bold or apply other text formatting. Doing that server side doesn't make sense as the CLI wouldn't want HTML tags for text formatting...

My suggestion going forward is that we simply skip this formatting, I think the messages are clear enough without putting the prefix in bold or similar and that we move to generating the message on the server side for simplicity.

The first question is thus, is it OK to merge these interpretation-keys into the dictSQL and is the above a suitable way?

Second question pertains to where we generate the actual text messages. Client side like today or should we move to server side?

I would also like to show a more complex query, in this case for mathing a tag. Searching for '#foo' will actually expand not only to match on tags=#foo but also on the inherited_tags attribute;
```
{
   'operator': 'or',
   'val2': {
      'operator': 'equals_any',
[...]

PR #786, opened 2015-05-22
cli: add customisable output columns

### LINKED ISSUE #684: Add CLI argument to specify output columns
While the CLI isn't meant for scripting (you should use Pynipap!!) there are cases when it is fairly handy to do a bit of simple scripting using bash, awk and friends. To simplify this further, we should have an argument to the nipap CLI which can be used to specify the output columns to print.

Further, we should have an argument to suppress column headers and similar that makes it more difficult to parse out the correct columns.

plajjan:
The output columns for 'nipap address list' can now be specified with
the -c/--columns paramter, which takes a comma separated list of column
names. To append columns to the default columns, prefix the list with a
"+".

Specify custom columns:

    kll@lingloi320 ~ $ nipap address list -c prefix
    Prefix
    ===============================
    1.0.0.0/8
      1.0.0.0/16
        1.0.0.0/24
          1.0.0.128/25
    2001:db8:1:1::/64

Append the free_addresses to the default set of columns:

    kll@lingloi320 ~ $ nipap address list -c +used_addresses,free_addresses
    VRF RT         Prefix                           #  Node   Order ID  Customer ID  Description  Used addresses  Free addresses
    ===================================================================================================================================
    -              1.0.0.0/8                      R #1 None   None      None         foobar       65536           16711680
    -                1.0.0.0/16                   R -  None   None      None         foobar       256             65280
    -                  1.0.0.0/24                 R -  None   None      None         foobar       128             128
    -                    1.0.0.128/25             A -  None   None      None         test         0               128
    -              2001:db8:1:1::/64              R -  None   None      None         test         0               18446744073709551616

In addition, the column width is now dynamic for the majority of
columns. The exceptions are columns that have a more "complicated"
value, like pool_name which is p.pool.name.

Fixes #684.

@garberg, I don't want you to merge this yet.. I'm just looking for
feedback. Please regard this as material for discussion.  The patch
feels kind of convoluted and I'm not particularly happy about the
code.. yet it does what it should :P

So after removing a bunch of lists in my recent PR for #781 I am now
introducing a list of columns, although this time on the client side..
but with reference to our offline discussion, do you see how using
pynipap on both client side and server could potentially save us a few
lines of code!?

I'm not sure if doing getattr in this fashion is a good idea or not...
can't access things like p.vrf.rt, or at least I don't know how!?

### DIFF
--- a/nipap-cli/nipap
+++ b/nipap-cli/nipap
@@ -55,6 +55,7 @@ if __name__ == '__main__':
 
     import argparse
     parser = argparse.ArgumentParser(description='NIPAP command-line client')
+    parser.add_argument("-c", "--columns", type=str, help="define columns")
     parser.add_argument("-f", "--force", action="store_true", help="disable interactive prompting of actions")
     parser.add_argument("--show-interpretation", action="store_true", help="show search interpretation")
     parser.add_argument("--version", action="store_true", help="display version information and exit")
--- a/nipap-cli/nipap
+++ b/nipap-cli/nipap
@@ -75,7 +75,7 @@ if __name__ == '__main__':
         sys.exit(1)
 
     # read configuration
-    cfg = ConfigParser.ConfigParser()
+    cfg = ConfigParser.ConfigParser({'prefix_list_columns': None })
     cfg.read(userrcfile)
     nipap_cli.nipap_cli.cfg = cfg
 
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -412,11 +412,51 @@ def list_prefix(arg, opts, shell_opts):
         }
     print "Searching for prefixes in %s..." % vrf_text
 
+    col_def = {
+            'added': { 'title': 'Added' },
+            'alarm_priority': { 'title': 'Alarm Prio' },
+            'authoritative_source': { 'title': 'Auth source' },
+            'children': { 'title': 'Children' },
+            'comment': { 'title': 'Comment' },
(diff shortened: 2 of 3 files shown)


fischa:
Would this include the possibility to exclude certain type of prefixes? E.g. for subnet planning you don't want to see host "prefixes". Or you just want to see assignements or reservations?

plajjan:
@fischa no it would not, this only allows you to modify the output columns, not the data retrieved. You are probably interested in something like #683 for that.

plajjan:
@garberg I've worked a bit more on this, making it a tad more dynamic.. I'm fairly happy with it at this point. What bugs me the most is probably just the size of the list_prefix function which is now up to 145 lines. Should probably be split up. Doing a generator instead of this for loop would be one thing. Anyway, I think that's something we can work separately one and disregard for this PR, right?

This only targets listing prefixes, so we might want to implement the same for VRFs and pools (not sure if we create separate PR for that or not though). Let me know what you think.

garberg:
Yeah, looks good to me! I agree that we could wait with the refactoring. IMHO the same goes for implementing the feature for VRFs and pools.

One thing that would be great to have already from day one and pretty simple to add would be the possibility to specify the default columns persistently in ~.nipaprc. What do you think, perhaps something we could add right now?

plajjan:
@garberg BAM! also fixed an issue similar to the one you did for the bulk rename script (i.e. not check my var before doing if var[0] == '+'...

Waddya think?

garberg:
Nice, let's merge!

PR #787, opened 2015-05-24
backend: rework attribute list for pool

### LINKED ISSUE #784: Rework pool spec/attr check
Do the equivalent of #667 but for pools.

plajjan:
This implements the equivalent of what was done in
93fbf8de6b598cd4e22568e7c3c59df04a60b5c8 but for pools. Instead of
spreading out a list of attributes in various places we keep one global
list which we reference. Neater code!

Fixes #784.


garberg:
Looks nice, merging!

PR #788, opened 2015-05-24
backend: rework attribute list for VRF

### LINKED ISSUE #785: Rework VRF spec/attr check
Do the equivalent of #667 but for VRF.

plajjan:
This implements the equivalent of what was done in
93fbf8de6b598cd4e22568e7c3c59df04a60b5c8 but for VRFs. Instead of
spreading out a list of attributes in various places we keep one global
list which we reference. Neater code!

Fixes #785.


garberg:
Looks nice, merging!

PR #795, opened 2015-05-28
authlib: Separate LDAP search credentials

### LINKED ISSUE #794: Separate LDAP search credentials
For some LDAP setups it's required to use separate credentials to perform the LDAP search for the user logging in. The search is performed after authentication to get the full name of the user and permissions (if used).

garberg:
Added functionality to allow NIPAP administrators to specify separate
LDAP credentials used when performing the LDAP search for the user
logging in. This is needed for LDAP setups where ordinary user accounts
are not permitted to search the LDAP database.

The search for the user logging in is performed after the user has been
authenticated to get full name and permissions (if used).

Fixes #794


plajjan:
Looking good!

PR #801, opened 2015-06-02
cli: fix prefix type guessing

### LINKED ISSUE #800: Incorrect prefix type guessing
So, this is kind of broken:
```
root@debian:~# nipap address list foobar
VRF            Prefix                       #  Node                Order          Customer       Description
============================================================================================================
-              10.0.0.0/16                A -  None                None           None           foobar
-              10.1.0.0/16                A -  None                None           None           foobar
root@debian:~# nipap address add prefix 10.0.0.0/8 type reservation description "foobar pappa"
WARNING: Parent prefix is of type 'assignment'. Automatically overriding specified type 'reservation' with type 'host' for new prefix.
Host 10.0.0.0/16 added to VRF 'default' [RT: -]: foobar pappa
root@debian:~#
```

plajjan:
The logic to automatically guess type of a new prefix failed when trying
to add a prefix larger than one already existing but with the same
network mask as the search to find the parent prefix would ignore the
prefix length. This is now fixed!

For example, 10.0.0.0/24 is an assignment and we are trying to add
10.0.0.0/8 as a reservation. The auto type guess would strip away the
prefix length, resulting in a search for "10.0.0.0" which the backend
interprets as 10.0.0.0/32 and that returns 10.0.0.0/24 as the parent
prefix. Since 10.0.0.0/24 is an assignment, the type would be changed to
"host" and the new prefix would be created with a /32 mask.

Now we check if there are bits set to the right of the mask and only
then do we assume a /32, otherwise we search with the prefix length.
Naturally 10.0.0.0/8 would not be contained by 10.0.0.0/24 so it can be
correctly created!

Fixes #800.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -557,12 +557,27 @@ def add_prefix(arg, opts, shell_opts):
         parent_prefix = args['from-prefix'][0]
         parent_op = 'equals'
     else:
-        parent_prefix = opts.get('prefix').split('/')[0]
-        parent_op = 'contains'
+        # If no prefix length is specified it is assumed to be a host and we do
+        # a search for prefixes that contains the specified prefix. The last
+        # entry will be the parent of the new prefix and we can look at it to
+        # determine type.
+        # If prefix length is specified (i.e. CIDR format) we check if prefix
+        # length equals max length in which case we assume a host prefix,
+        # otherwise we search for the network using an equal match and by
+        # zeroing out bits in the host part.
+        if len(opts.get('prefix').split("/")) == 2:
+            ip = IPy.IP(opts.get('prefix').split("/")[0])
+            plen = int(opts.get('prefix').split("/")[1])
+            if ip.version() == 4 and plen == 32 or ip.version() == 6 and plen == 128:
+                parent_prefix = str(ip)
+                parent_op = 'contains'
+            else:
+                parent_prefix = str(IPy.IP(opts.get('prefix'), make_net=True))
+                parent_op = 'equals'
+        else:
+            parent_prefix = opts.get('prefix')
+            parent_op = 'contains'
 
-    # prefix must be a CIDR network, ie no bits set in host part, so we
-    # remove the prefix length part of the prefix as then the backend will
-    # assume all bits being set
     auto_type_query = {
             'val1': {
                 'val1'      : 'prefix',
--- a/tests/nipaptest.py
+++ b/tests/nipaptest.py
@@ -1575,7 +1575,6 @@ class TestPrefixLastModified(unittest.TestCase):
 class TestCli(unittest.TestCase):
     """ CLI tests
     """
-
     def test_extra_args(self):
         """ Extra arg should raise exception
         """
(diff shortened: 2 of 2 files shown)


nipap-cli/nipap_cli/nipap_cli.py
```diff
@@ -557,12 +557,17 @@ def add_prefix(arg, opts, shell_opts):
...
         parent_op = 'contains'
 
-    # prefix must be a CIDR network, ie no bits set in host part, so we
+    # If bits are set in the host part we assume a host is being added and
     # remove the prefix length part of the prefix as then the backend will
     # assume all bits being set
+    try:
+        _ = IPy.IP(parent_prefix)
+    except:
+        parent_prefix = opts.get('prefix').split("/")[0]
```
garberg: If we, from some reason, would end up here when assigning a prefix from a prefix (which seems pretty unrealistic), the option 'prefix' doesn't exist (and is wrong).

nipap-cli/nipap_cli/nipap_cli.py
```diff
@@ -557,12 +557,17 @@ def add_prefix(arg, opts, shell_opts):
...
+        parent_prefix = opts.get('prefix')
         parent_op = 'contains'
 
-    # prefix must be a CIDR network, ie no bits set in host part, so we
+    # If bits are set in the host part we assume a host is being added and
     # remove the prefix length part of the prefix as then the backend will
     # assume all bits being set
+    try:
+        _ = IPy.IP(parent_prefix)
+    except:
```
garberg: Catch ValueError explicitly, better safe than sorry? :smile:

nipap-cli/nipap_cli/nipap_cli.py
plajjan: Right, I better just put this in the above else clause!

nipap-cli/nipap_cli/nipap_cli.py
plajjan: Bah :P

garberg:
Hm.. there are some corner-cases with prefix lengths 31 and 127, right? What happens when getting the first address in a /31 och /127?

Say we have these addresses:
```
R 192.0.2.0/24
A   192.0.2.2/31
```
When allocating a host address in 192.0.2.2/31 specifying the prefix to allocate as 192.0.2.2/31, no bits are set to the right of the mask and the parent search will return the parent of 192.0.2.2/31, the reservation 192.0.2.0/24 and not the assignment. Then the prefix type logic is then bypassed... right? IMHO that's OK for /31 and /127.

fischa:
Isn't there another corner case for loopbacks or similar approaches where the network address is used because single IPs are picked for services?
```
# nipap address list 192.0.2.0/24
VRF            Prefix                          #  Node                Order          Customer       Description
===============================================================================================================
-              192.0.2.0/24                  A -  None                None           None           loopbacks
-                192.0.2.0/24                H -  None                None           None           router1
```
Is this covered as well?

hallinde:
Had this error occur moments ago, how is it going along? 
My version of the story looked like this, bit obfuscated.


~$ nipap address add vrf_rt Y:Z prefix 10.244.0.0/16 type reservation
WARNING: Parent prefix is of type 'assignment'. Automatically overriding specified type 'reservation' with type 'host' for new prefix.
Host 10.244.0.0/29 added to VRF 'secrets' [RT: Y:Z]: None

~$ nipap address add vrf_rt Y:Z prefix 10.244.0.0/16 type reservation
WARNING: Parent prefix is of type 'assignment'. Automatically overriding specified type 'reservation' with type 'host' for new prefix.
Could not add prefix to NIPAP: Duplicate value for 'prefix', the value '74, 10.244.0.0' is already in use.

~$ nipap address add vrf_rt Y:Z prefix 10.242.0.0/29 type assignment
Network 10.242.0.0/29 added to VRF 'secrets' [RT: Y:Z]: None

~$ nipap address add vrf_rt Y:Z prefix 10.242.0.0/16 type reservation
WARNING: Parent prefix is of type 'assignment'. Automatically overriding specified type 'reservation' with type 'host' for new prefix.
Host 10.242.0.0/29 added to VRF 'secrets' [RT: Y:Z]: None

plajjan:
@hallinde if I'm not mistaken you are on an older version as well and have some work to do with integration of other systems before you can upgrade so even if we did apply a patch here I'm not sure you could use it.

Anyway, I should test this with the given scenarios, like loopbacks et al. I'm not sure if it works with my patch or not. Will try to wrap it up in coming week.

plajjan:
@fischa btw, you shouldn't register your loopbacks as hosts in an assignment. Think of the assignment to match the natural netmask of the prefix. You configure loopbacks as /32 or /128, so you should register each loopback as an assignment inside a reservation and not like hosts inside an assignment. I know this can be confusing but just think that the assignment prefix length should always match up to what you see in your routing table.

plajjan:
Ok, I took another stab at this. Rewrote my logic from scratch and added a bunch of unittests. I'm pretty certain this now does the right thing (tm).

@garberg please have a look!

garberg:
Looks good to me, nice with tests as well. Merging!

PR #809, opened 2015-06-22
Unicode support for whoisd

### LINKED ISSUE #808: whoisd doesn't work with entries using umlaut/accent characters
While using special characters like German Umlauts (äöüß) on both the web interface as well as the command line tool works fine, using the whois interface fails for any entry using them. And that not only when trying to search for an umlaut string itself, but also a prefix containing such a character in its description. e.g.:

root@ipam1:~# nipap address list 192.168.10.128/29
VRF            Prefix                                     #  Node                Order          Customer       Description
==========================================================================================================================
-              192.168.0.0/16                           R -  None                None           None           NETHINKS PA /16
-                192.168.0.0/20                         R -  None                None           None           None
-                  192.168.10.0/24                      R -  None                None           None           None
-                    192.168.10.128/29                  A -  None                None           15641          Günther (aDSL)
-                      192.168.10.129/29                H -  None                None           None           None
root@ipam1:~# nipap address list Günther
VRF            Prefix                                   #  Node                Order          Customer       Description
========================================================================================================================
-              192.168.0.0/16                         R -  None                None           None           NETHINKS PA /16
-                192.168.0.0/20                       R -  None                None           None           None
-                  192.168.10.0/24                    R -  None                None           None           None
-                    192.168.10.128/29                A -  None                None           15641          Günther (aDSL)
-                192.168.32.0/19                      R -  None                None           None           None
-                  192.168.41.0/24                    R -  None                None           None           None
-                    192.168.41.144/28                A -  None                None           15641          Günther EVPN 4M

But:
root@ipam1:~# whois -hlocalhost 192.168.10.128/29
root@ipam1:~# whois -hlocalhost 192.168.10.128/24
root@ipam1:~# whois -hlocalhost 192.168.10.0/24
root@ipam1:~# whois -hlocalhost 192.168.10.0
% This is the NIPAP whois query service.

inetnum:             192.168.0.0/16
vrf-name:            default
descr:               NETHINKS PA /16
country:             DE
monitor:             False

inetnum:             192.168.0.0/20
vrf-name:            default
monitor:             False

inetnum:             192.168.10.0/24
vrf-name:            default
monitor:             False

inetnum:             192.168.10.0/29
vrf-name:            default
descr:               EVPN XXX
country:             DE
monitor:             False

% This query was served by the NIPAP whois server version 0.27.3

Please note that not only the affected entry isn't output, but rather ALL the lines, so I assume the whois daemon seems to have some sort of internal exception ...

plajjan:
This adds Unicode support for whoisd by decoding requests using IDNA and/or UTF-8.

Responses are now sent in UTF-8.

Not sure if we should merge this on 0.28 or wait for 0.29.

Fixes #808.

### DIFF
--- a/whoisd/nipap-whoisd
+++ b/whoisd/nipap-whoisd
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import unicode_literals
+
 import fcntl
 import os
 import re
@@ -77,6 +79,10 @@ class WhoisServer(SocketServer.BaseRequestHandler):
                 break
 
         query = query.strip()
+        try:
+            query = query.decode("idna")
+        except:
+            pass
 
         # respond!
         total_hits = 0
@@ -128,7 +134,7 @@ class WhoisServer(SocketServer.BaseRequestHandler):
             response += '%\n'
 
         response += "\n%% This query was served by the NIPAP whois server version %s\n" % nipap_whoisd.__version__
-        self.request.send(response)
+        self.request.send(response.encode("utf-8"))
 
 
 
--- a/whoisd/nipap-whoisd
+++ b/whoisd/nipap-whoisd
@@ -66,19 +66,13 @@ class ForkingTCPServer(ForkingMixIn, TCPServer):
         SocketServer.TCPServer.server_bind(self)
 
 
-class WhoisServer(SocketServer.BaseRequestHandler):
+class WhoisServer(SocketServer.StreamRequestHandler):
     def handle(self):
         """ Called for every connection
         """
         # receive data
-        query = ''
-        while True:
-            data = self.request.recv(1024)
-            query += data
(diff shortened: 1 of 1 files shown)


plajjan:
We could add a config option to specify output encoding if we feel that strict standards compliance is a good thing.

@garberg also read #808 for some discussion on the topic :)

garberg:
Looks good to me. The readline() instead of read(1024) will break multi-line queries, but I don't really see that as an issue for a whois daemon...

In my eyes this is more of a "feature" and should then perhaps be merged into 0.29 instead of 0.28. But on the other hand, perhaps it's time for 0.29? :)

plajjan:
@garberg you're right. We'll hold on to this for 0.29!

garberg:
... and now it's here, merging!

PR #810, opened 2015-06-22
pynipap: remove xmlrpc dependency

### LINKED ISSUE #805: Remove xmlrpc requirement for pynipap
The setup.py file for pynipap list xmlrpclib as a requirement but that is part of the standard Python library so I don't think it is required to list it there at all!? Shouldn't we just remove it?

plajjan:
xmlrpclib is shipped with Python so I don't think we need to declare a
dependency on it.

Fixes #805.

### DIFF
--- a/pynipap/setup.py
+++ b/pynipap/setup.py
@@ -5,10 +5,6 @@ import sys
 
 import pynipap
 
-requirements = []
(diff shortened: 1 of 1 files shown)


garberg:
Makes sense...

PR #812, opened 2015-06-25
Update VRF filter when VRF is removed

garberg:
Make sure the VRF filter list is updated when a VRF in the filter list is removed from NIPAP.


plajjan:
Looks good - merging!

PR #813, opened 2015-06-26
utilities: Added script to bulk rename prefixes

garberg:
Added a script which can be used to "rename" prefixes in NIPAP, that is change description and node, in bulk. When executed, the script will display a list of the prefixes to change and what they will be changed to as well as let the user select what prefixes to perform the operation on.  Either all, a few selected prefixes or all prefixes EXCEPT the selected prefixes can be chosen.

### DIFF
--- a/utilities/bulk-string-replace.py
+++ b/utilities/bulk-string-replace.py
@@ -1,0 +1,187 @@
+#! /usr/bin/env python
+""" Tool to replace strings in NIPAP
+
+    This tool is used to replace strings in NIPAP, helpful when for example
+    moving many prefixes from a node to another. Instead of manually update
+    prefix nodes and descriptions, this tool replaces all instances of a string
+    with another.
+    
+    Currently the following attributes are checked for strings to replace:
+    * node
+    * description
+
+    Usage: bulk-string-replace.py pattern replacement
+
+    When the script is executed, it begins by fetching all prefixes from NIPAP
+    which matches the pattern. It then iterates over them and prints out a
+    "diff", showing what lines matches and what they would be changed to. Above
+    each entry a number is displayed which identifies the entry.
+
+    Then the user is asked to select what changes to perform. The selection is
+    done by entering the numbers identifying the entries to change. Three types
+    of input is supported:
+    * The string 'all' - perform all listed replacements
+    * A comma separated list of numbers (eg. 5,7,12,37).
+      Replacement will be performed on the listed entries.
+    * A comma separated list of numbers prefixed with an exclamation mark (!, eg
+      !5,7,12,37). Replacement will be performed on all entries EXCEPT the
+      listed entries.
+"""
+
+import ConfigParser
+import os
+import time
+import re
+import sys
+
+from pynipap import Prefix, Pool, VRF
+import pynipap
+
+BATCH_SIZE = 100
+
+COLOR_RESET = "\x1b[0m"
+COLOR_RED = "\x1b[91m"
+COLOR_GREEN = "\x1b[92m"
+
+def replace(pattern, replacement):
+
+    # Fetch prefixes matching the string to replace
+    print "Fetching prefixes from NIPAP... ",
+    sys.stdout.flush()
+    n = 1
+    prefix_list = []
+    t0 = time.time()
+    query = {
+        'operator': 'or',
+        'val1': {
+            'operator': 'regex_match',
+            'val1': 'description',
+            'val2': pattern
+        },
+        'val2': {
+            'operator': 'regex_match',
+            'val1': 'node',
+            'val2': pattern
+        }
+    }
+    full_result = Prefix.search(query, { 'parents_depth': -1, 'max_result': BATCH_SIZE })
+    prefix_result = full_result['result']
+    prefix_list += prefix_result
+    print len(prefix_list), 
+    sys.stdout.flush()
+    while len(prefix_result) == 100:
+        full_result = Prefix.smart_search(pattern, { 'parents_depth': -1, 'max_result': BATCH_SIZE, 'offset': n * BATCH_SIZE })
+        prefix_result = full_result['result']
+        prefix_list += prefix_result
+        print len(prefix_list), 
+        sys.stdout.flush()
+        n += 1
+
+    t1 = time.time()
+    print " done in %.1f seconds" % (t1 - t0)
+
+    # Display list
+    print_pattern = "%-2s%-14s%-2s%-30s%-20s%s"
+    print "\n\nPrefixes to change:"
+    print print_pattern % ("", "VRF", "", "Prefix", "Node", "Description")
+    i_match = 0
+    for i, prefix in enumerate(prefix_list):
+        if prefix.match:
+            print COLOR_RESET,
+            print " -- %d --" % i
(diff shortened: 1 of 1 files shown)


plajjan:
I don't think "rename" is a good name for this script although I do understand where it originates from. Prefixes don't really have names.. this is more of pattern replacement, so can we rename the script accordingly?

Maybe one day we should consider integrating this in the CLI but I think scripts like this is good in the meantime.

garberg:
What about bulk-update?

utilities/bulk-rename.py
```diff
@@ -0,0 +1,179 @@
+#! /usr/bin/env python
+""" Tool to replace strings in NIPAP
+
+    This tool is used to replace strings in NIPAP, helpful when for example
+    moving many prefixes from a node to another. Instead of manually update
+    prefix nodes and descriptions, this tool replaces all instances of a string
+    with another.
+    
+    Currently the foolowing attributes are checked for strings to replace:
```
plajjan: fool ;)

utilities/bulk-rename.py
```diff
@@ -0,0 +1,179 @@
...
+            'val1': 'node',
+            'val2': pattern
+        }
+    }
+    full_result = Prefix.search(query, { 'parents_depth': -1, 'max_result': BATCH_SIZE })
+    prefix_result = full_result['result']
+    prefix_list += prefix_result
+    print len(prefix_list), 
+    sys.stdout.flush()
+    while len(prefix_result) == 100:
```
plajjan: I'm getting tired of this pattern. We should build a generator out of it so it's more natural to loop over prefixes.. It's not really related to your PR though ;)

garberg:
OK, fixed type and renamed the script. I noticed that I myself referred to the function of the script as string replacement, so I went with bulk-string-replace.py.

utilities/bulk-string-replace.py
```diff
@@ -0,0 +1,179 @@
...
+    print_pattern = "%-2s%-14s%-2s%-30s%-20s%s"
+    print "\n\nPrefixes to change:"
+    print print_pattern % ("", "VRF", "", "Prefix", "Node", "Description")
+    i_match = 0
+    for i, prefix in enumerate(prefix_list):
+        if prefix.match:
+            if use_colors:
+                print "\x1b[0m",
+            print " -- %d --" % i
+            color = "\x1b[91m"
```
plajjan: Does this really honour use_colors? You use it just above but here we set the color variable which is used on line 96 to print stuff. I don't know color codes by heart. Aren't there names for these in some lib or can we define them somewhere instead?

utilities/bulk-string-replace.py
```diff
@@ -0,0 +1,179 @@
...
+            prefix.vrf.rt,
+            prefix.type[0].upper(),
+            (("  " * prefix.indent) + prefix.display_prefix)[:min([ len(prefix.display_prefix) + 2*prefix.indent, 30 ])],
+            (prefix.node or '')[:min([ len(prefix.node or ''), 20 ])],
+            (prefix.description or '')[:min([ len(prefix.description or ''), 900 ])]
+        )
+        if prefix.match:
+            new_prefix_node = re.sub(pattern, replacement, (prefix.node or ''), flags=re.IGNORECASE)
+            new_prefix_desc = re.sub(pattern, replacement, (prefix.description or ''), flags=re.IGNORECASE)
+            print ("\x1b[92m" + print_pattern) % (
```
plajjan: Same thing about colours here, is use_color honoured or not?

utilities/bulk-string-replace.py
```diff
@@ -0,0 +1,179 @@
...
+    print "\x1b[0m",
+
+    # Perform action?
+    print "Select replacements to perform"
+    print "Enter comma-separated selection (eg. 5,7,10) or \"all\" for all prefixes."
+    print "Prefix list with ! to invert selection (eg !5,7,10 to perform operation on all except the entered prefixes)"
+    inp = raw_input("Selection: ").strip()
+
+    invert = False
+    if inp[0] == "!":
```
plajjan: I guess this fails if inp is empty, i.e. user just pressed enter.

utilities/bulk-string-replace.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+    # reset colors
+    print "\x1b[0m",
+
+    # Perform action?
+    print "Select replacements to perform"
+    print "Enter comma-separated selection (eg. 5,7,10) or \"all\" for all prefixes."
+    print "Prefix list with ! to invert selection (eg !5,7,10 to perform operation on all except the entered prefixes)"
+    inp = raw_input("Selection: ").strip()
```
plajjan: There doesn't appear to be a way to cancel any replacement!?

^C ?

Also, I think "inp" is a silly abbreviation of input ;)

utilities/bulk-string-replace.py
```diff
@@ -0,0 +1,179 @@
...
+    parser.add_argument('--password', help="password")
+    parser.add_argument('--host', help="NIPAP backend host")
+    parser.add_argument('--port', help="NIPAP backend port")
+    args = parser.parse_args()
+
+    # Set up NIPAP
+    auth_uri = "%s:%s@" % (args.username or cfg.get('global', 'username'),
+            args.password or cfg.get('global', 'password'))
+
+    xmlrpc_uri = "http://%(auth_uri)s%(host)s:%(port)s" % {
```
plajjan: guess this fails if I put in an IPv6 address here ;)

Most other scripts fail in a similar manner but perhaps we should start improving this. Can we always do [%(host)s]? I guess it works for IPv4 too, right?

garberg: Yup, seems to work, I'll update. This was taken from boilerplate.py, also it should be updated.

utilities/bulk-string-replace.py
garberg: Yeah, ^C was the intention (even tough it throws a nasty traceback in the user's face).

I wanted to avoid input as that's the name of a built-in function, see https://docs.python.org/2/library/functions.html#input. I could probably have chosen a better name though.

utilities/bulk-string-replace.py
garberg: Yup :smile:

garberg:
I think you vastly overestimate the ambition of this script :stuck_out_tongue: 

But sure, I'll do some updates. There seems to be a few libraries which assist with colors, for example https://pypi.python.org/pypi/colorama. Regarding use_colors, I guess that could be removed entirely. As it's an interactive script it'll probably be very rare that it'll write to anything other than a TTY.

plajjan:
@garberg haha, perhaps. I have no major issue merging it as is really.. it's better having the script in its current state than no script at all but just fixing the host thing for IPv6 and the empty input should be easy and then we're good to go.

I agree with you on the colors, I think you can assume color terminal. If that proves to be wrong then.. well, too bad :P

garberg:
Okay, some changes! Handle empty and invalid input, remove TTY verification, fixed issue with IPv6 hosts and replaced magic color characters with helpfully named variables. I chose not to use any terminal color library to avoid adding dependencies.

PR #816, opened 2015-06-30
Add generic export script based on template

### LINKED ISSUE #815: Add generic export facility to file
Add a generic export to file script that uses a template file. This could be used to produce various configuration files based on NIPAP data.

plajjan:
This adds a generic export script that can be used for producing various
configuration files based on data from NIPAP.

In addition to the script, I have provided a basic template for ISC DHCP
config (not tested).

Fixes #815.

### DIFF
--- a/utilities/export-template.py
+++ b/utilities/export-template.py
@@ -1,0 +1,105 @@
+#!/usr/bin/python
+""" This is a generic file exporter for NIPAP
+
+    It fetches prefixes based on a provided query string and feeds that into a
+    template that can be used to produce a configuration file for another
+    program, such as ISC DHCP.
+
+    It will read .nipaprc per default or rely on command line options for
+    connection settings to the backend.
+"""
+
+from __future__ import print_function
+
+import ConfigParser
+import os
+# use local pynipap, useful if we are developing
+import sys
+sys.path.append('../pynipap')
+from pynipap import Prefix, Pool, VRF
+import pynipap
+
+import IPy
+import jinja2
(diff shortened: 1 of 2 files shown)


crlsgms:
that would be awesome. Ill test asap to populate my nipap with my current ipam data, ill keep posted

utilities/export-template.py
```diff
@@ -0,0 +1,105 @@
...
+    cfg = ConfigParser.ConfigParser()
+    cfg.read(os.path.expanduser('~/.nipaprc'))
+
+    import optparse
+    parser = optparse.OptionParser()
+    # standard options to specify nipapd connection
+    parser.add_option('--username', default='', help="Username")
+    parser.add_option('--password', default='', help="Password")
+    parser.add_option('--host', help="NIPAP backend host")
+    parser.add_option('--port', default=1337, help="NIPAP backend port")
```
garberg: The default-argument should probably be removed here as it leads to the port never being taken from ~/.nipaprc. As ```options.port``` is always set due to the default-value, ```options.port or cfg.get('global', 'port')``` at line 90 will never yield the value from cfg. Originally an issue with boilerplate.py, I'll create an issue for that as well.

garberg:
Nice addition! Just fix the default-issue I mentioned in the diff comment and I'll happily merge this :smile:

plajjan:
I've removed the default for username, password and port now!

garberg:
Cool, merging!

PR #822, opened 2015-07-13
Remove default arguments in boilerplate

### LINKED ISSUE #821: Remove optparse default values in boilerplate.py
The default values for username, password and port passed to optparse in boilerplate.py essentially disables reading these values from ~/.cloginrc. Due to the default values the ```options```-object generated by optparse will always have values set and thus will the statements ```options.xxx or cfg.get('xxx')``` always evaluate to ```options.xxx```.

plajjan:
The default arguments in the argument parser prevents the use of the
values from the configuration file.

Fixes #821.

### DIFF
--- a/utilities/boilerplate.py
+++ b/utilities/boilerplate.py
@@ -26,10 +26,10 @@ if __name__ == '__main__':
 
     import optparse
     parser = optparse.OptionParser()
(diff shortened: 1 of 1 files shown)


garberg:
:thumbsup:

PR #823, opened 2015-07-15
Enhance smart parsing

plajjan:
This completely replaces the code and logic behind how we do smart parsing of search queries. Instead of relying on shlex and dumb parsing of individual tokens we now use pyparsing to tokenise the input string and work on the AST that it produces. I'm fairly happy with this approach. It seems quite resilient and is able to parse fairly complex queries, including things like correctly understanding quoting, including escapes quotes, parentheses (and quoted parentheses that shouldn't be parsed).

Queries like "vlan=100" (or "vlan = 100") are now possible, as are using OR to join together expressions in the query. There is still an implicit AND between the keywords in the query where there is no explicit operator.

The test suite runs for the new parser with minimal modifications, I've only changed the expected text output when dealing with an unclosed quote. I have added some test to cover a bit more complex queries but we can (and should) certainly add more tests.

The --show-interpretation of the CLI is borked. I think this comes from the dictSQL being structured differently. I haven't looked into it too much yet but I think the difference is that when we build a "deep" query and joining it together by AND-ops we used to build the structure in val1 and now it is in val2, if you get what I mean ;) This confuses the query printing of the CLI so the output is kind of messed up.

Apart from the CLI interpretation, which needs to be fixed before this is merged, I believe this to be complete.


plajjan:
Heads up to @fischa. You asked in (#786) about possibility for more flexible queries.. this is it! =)
Feedback is welcome if you want to try it out.

fischa:
Thanks :) Not sure if I can make it in a timely fashion but I will have a look.

nipap/nipap/smart_parsing.py
```diff
@@ -0,0 +1,744 @@
...
+            dss = self._string_to_dictsql(ParseResults('', 'word'))
+
+        # return the final composed stack of dictsql expressions
+        return dss
+
+
+    def _string_to_dictsql(self, string):
+        """ Do magic matching of single words or quoted string
+        """
+        raise NotImplemented()
```
garberg: Should be NotImplementedError, no?

nipap/nipap/smart_parsing.py
```diff
@@ -0,0 +1,744 @@
...
+        comp_word = Word(alphanums + "-./_")
+        word = Word(alphanums + "-./_").setResultsName('word')
+        # numbers
+        comp_number = Word(nums)
+        number = Word(nums).setResultsName('number')
+
+        # IPv4 address
+        ipv4_oct = Regex("((2(5[0-5]|[0-4][0-9])|[01]?[0-9][0-9]?))")
+        comp_ipv4_address = Combine(ipv4_oct + ('.' + ipv4_oct*3))
+        ipv4_address = Combine(ipv4_oct + ('.' + ipv4_oct*3)).setResultsName('ipv4_address')
```
garberg: What about IPv6? :smile:

garberg:
Okay, I've read through the PR briefly and made a couple of comments without really understanding the code in detail. Nice work! Also nice to have the exceptions and parsing moved to separate modules, the backend module is pretty large...

nipap/nipap/smart_parsing.py
plajjan: Yupp, you're right. I'll fix.

nipap/nipap/smart_parsing.py
plajjan: So there are two groups here; "comp_ipv4_address" and "ipv4_address". They match identical strings but the later is named (setResultsName). If they are named or not will give different results (having lots of named components means a deep nested dict is returned) when used to match, which is why I also have comp_ipv4_address, "comp" is a short for component and it is meant to be used as a component in a larger matching, in this case it is used for vrf_rt, which you can see on line 106. I don't care about  VRF RT "1.2.3.4:789" being built out of an ipv4_address : number, I just want to see it as one.

"ipv4_address" isn't actually used at all. I wanted to use it but had some issues and decided it was easier to leave it out for now. We still do interpret "1.2.3.0/24" as an IPv4 prefix, but pyparsing thinks it is a text and it is first on line 471 (in the PrefixSmartParsing) that we do parsing of this text string to identify IPv4/IPv6 prefixes. As you might notice, it is the code from our previous smart_parsing function that has been moved over.

garberg:
OK, merging!

PR #824, opened 2015-07-15
sql: fix parent_prefix selection

### LINKED ISSUE #817: unable to shrink assignment prefix
when trying to shrink a assignment prefix, I´ll get this error: Parent prefix (10.0.3.0/24) is of type assignment but must be of type 'reservation'

but this prefix is the prefix I´m trying to shrink, it seems the check checks self, but it should ignore self

it has been tested on 0.27.4 and 0.28.2

plajjan:
If a prefix is shrunk we have to filter it out in the before trigger to
avoid selecting its old self as its parent_prefix. This is perfectly
natural as the database before trigger sees the data before the update
has been applied.

This fixes the scenario when trying to shrink an assignment prefix.
Before it would select itself as parent and when running the consistency
checks on that it would refuse the update as an assignment cannot have
an assignment (its old self) as a parent.

The prefix statistics were written under this false assumption, so with
the correct parent_prefix it would generate incorrect statistics which
I've solved by adding an explicit path for when shrinking prefixes, i.e.
NEW is smaller than OLD but indent level is the same.

This fixes #817.

### DIFF
--- a/nipap/sql/triggers.plsql
+++ b/nipap/sql/triggers.plsql
@@ -135,7 +135,12 @@ BEGIN
 		i_max_pref_len := 128;
 	END IF;
 	-- contains the parent prefix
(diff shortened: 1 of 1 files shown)


garberg:
Looks good to me, merging!

PR #827, opened 2015-07-21
Update nipap.css

hallinde:
More sensible css to fit at least in 1680x1050 height resolution Chrome.
http://oi59.tinypic.com/28rilc9.jpg
left image is this change
right is original


plajjan:
I suppose the goal here is to fit all fields onto the screen to avoid scrolling. Overall this is tricky and there's no permanent solution. We need to support many different resolutions and there's just no way we can make guarantees on fitting everything on screen without scrolling. However, I like making things a bit more dense so I'll merge this. Hope it doesn't mess up any other pages (I haven't checked).

PR #830, opened 2015-07-23
Add requirements.txt for nipap-www

plajjan:
This adds a requirements.txt for nipap-www. I haven't actually tried it
out and it is possible that we should pin more dependencies to specific
version numbers but I think this should work for now.

Looking forward to feedback from whomever might be using this.

Fixes #627


plajjan:
Bah, closing this shite. I pushed the wrong file :/

PR #831, opened 2015-07-26
Add references to backend functions in pynipap docs

garberg:
Allright, I managed to bust up the original PR for this, #758, so here's a new one. I've added explanations regarding the mapping of functions in pynipap to backend functions to each reference. During offline discussions we debated whether the references should be to the XML-RPC functions or all the way to the backend functions, but from what I remember we agreed on referring to the backend (as that's where the information of interest actually is available), even though there is an XML-RPC layer in between.

Original message:
To avoid duplicating information the pynipap documentation is very brief as most data is passed straight to the backend anyway. This PR tries to make this more clear by adding references to the corresponding backend function to each of the pynipap class methods which involve backend operations.

### DIFF
--- a/docs/sphinx/conf.py
+++ b/docs/sphinx/conf.py
@@ -217,7 +217,9 @@ if on_rtd:
             else:
                 return Mock()
 
-    MOCK_MODULES = ['ldap', 'IPy', 'psycopg2.extras', 'psycopg2']
+    MOCK_MODULES = [
+        'ldap', 'IPy', 'psycopg2.extras', 'psycopg2', 'pyparsing'
+    ]
 
     for mod_name in MOCK_MODULES:
         sys.modules[mod_name] = Mock()
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -1189,6 +1189,10 @@ class Nipap:
             Add a VRF based on the values stored in the `attr` dict.
 
             Returns a dict describing the VRF which was added.
+
+            This is the documentation of the internal backend function. It's
+            exposed over XML-RPC, please also see the XML-RPC documentation for
+            :py:func:`nipap.xmlrpc.NipapXMLRPC.add_vrf` for full understanding.
         """
 
         self._logger.debug("add_vrf called; attr: %s" % str(attr))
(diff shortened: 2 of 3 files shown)


garberg:
Oh, and the docs generated from this branch can be seen at http://nipap-garberg.readthedocs.org/en/712-update_pynipap_docs/pynipap.html.

plajjan:
From #758:

"In addition, in the backend module, perhaps we should have a section explaining the XML-RPC API and how it wraps around the backend and for every backend function we have a reference saying something along the lines of "This is the documentation of the internal backend function. It is exposed over XML-RPC, please also see the documentation of the XML-RPC function for full understanding."."

Ie, I'd like to see a notice in the backend function that there is a XML-RPC wrapper around it. For example, see https://github.com/SpriteLink/NIPAP/issues/750 where the user was confused over the backend vs xml-rpc argument discrepancy. Or should that be another PR?

plajjan:
Apart from that I think it looks good!

plajjan:
Oh, and can we have a link in the menu to each class (or each function)? Right now it just says "Classes" which isn't very easy to understand or navigate...

garberg:
My thought was to leave the reference to XML-RPC in the backend to another PR, but what the heck. I might just as well include it here.

I'll see if we can get links to the classes i the menu as well.

plajjan:
Ok, trying to brush this off so we can get moving (this PR has celebrated 1 year!!!! :/)

Seems readthedocs doesn't build properly - compare sections in the menu for http://nipap-garberg.readthedocs.org/en/712-update_pynipap_docs/authlib.html with http://nipap.readthedocs.org/en/v0.28.4/nipap.html  guess we need to fix that first but @garberg feel free to rebase this on master as well :)

garberg:
OK, problem solved and rebased on current master. I looked through the result briefly, looks good to me. Should be ready to go :smile:

What's implemented is references:
* pynipap -> backend
* Backend -> XML-RPC

What's missing is reference:
* XML-RPC -> backed

... but let't create a separate issue for that!

plajjan:
Ok, I still think I would prefer pynipap -> XML-RPC -> backend mapping but we can discuss that some other time over a beer or whatnot ;)

PR #840, opened 2015-09-16
Update nipap_cli.py

hallinde:
Previously gave me and error and this seemed the best way to encode äöå at the time. 

Error message that I encountered
```
Recursively deleting 172.27.6.0/24 in VRF 'hemligt nät' [RT: 1337:210] will delete the following prefixes:
  172.27.6.0/24               R  None                None           None         back2back 
    172.27.6.0/29             A  None                None           None             back2back
      172.27.6.3/29           H  None                None           None           Gateway
    172.27.6.8/29             A  None                None           None             back2back
      172.27.6.11/29          H  None                None           None           Gateway
    172.27.6.16/29            A  None                None           None             back2back
      172.27.6.19/29          H  None                None           None           Gateway
    172.27.6.24/29            A  None                None           None             back2back
      172.27.6.27/29          H  None                None           None           Gateway
    172.27.6.32/29            A  None                None           None             back2back
      172.27.6.35/29          H  None                None           None           Gateway
.. and 54 other prefixes
  File "/usr/bin/nipap", line 122, in <module>
    cmd.exe(cmd.arg, cmd.exe_options, args)
  File "/usr/lib/python2.7/dist-packages/nipap_cli/nipap_cli.py", line 1220, in remove_prefix
    vrf_format(vrf)))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 82: ordinal not in range(128)
```

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -1295,7 +1295,7 @@ def remove_prefix(arg, opts, shell_opts):
                 if len(auth_src) == 0:
                     # Simple case; all prefixes were added from NIPAP
                     res = raw_input("Do you really want to recursively remove %s prefixes in %s? [y/N]: " % (len(pres['result']),
-                                vrf_format(vrf)))
+                                vrf_format(vrf).encode("utf-8")))
 
                     if res.lower() in [ 'y', 'yes' ]:
                         remove_confirmed = True
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -67,7 +67,7 @@ def setup_connection():
 
 
 def vrf_format(vrf):
-    return "VRF '%s' [RT: %s]" % (vrf.name, vrf.rt or '-')
+    return "VRF '%s' [RT: %s]" % (vrf.name.encode("utf-8"), vrf.rt or '-')
 
 
 def get_pool(arg = None, opts = None, abort = False):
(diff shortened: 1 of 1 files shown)


plajjan:
Thank you for your patch, however, I don't believe this is the proper solution. The output should already be UTF-8. I can see that you get UTF-8 output on the first line but it fails later on. Line 1220 is using raw_input - so it seems to be different than normal print.

Can you confirm your environment is properly setup for UTF-8?

Also, just as feedback on the actual PR (not the code content) you should try to squash incorrect commits (use git rebase -i HEAD~2 and select appropriate actions), see https://github.com/SpriteLink/NIPAP/blob/master/CONTRIBUTING.md for guidelines. The commit message isn't very helpful - you should try to describe what change you are doing.

I'll look into this to reproduce and see if we can come up with a more generic fix :)

hallinde:
Export declares,
declare -x LANG="en_US.UTF-8"
declare -x LANGUAGE="en_US:en"

Yeah, I still have some things to learn about GITting, a lot actually :+1:. Sorry about the added hassle

plajjan:
I did a quick test and I can reproduce this in my environment which is properly setup for UTF-8 so there is an issue.

Normally we rely on Python to do the right thing:tm: when it comes to encoding but there are exceptions and one such instance when we are being piped to another process. It will default to ASCII but we force it to UTF-8 because that probably makes more sense, that happens here https://github.com/SpriteLink/NIPAP/blob/master/nipap-cli/nipap#L23

As a test I tried using codecs.getwriter to force UTF-8, not just when we are using a pipe but all the time, which apparently fixes this issue. I don't think it's a good idea to force UTF-8 output in case someone is trying to use this on some old arcane terminal. Funny thing is how sys.stdout.encoding is "UTF-8" before AND after I set the value - i.e. I can't see a difference, yet it makes a difference.

I guess one way would be to check if encoding is UTF-8, then we set it to UTF-8, and this problem should go away.. for everyone using an UTF-8 terminal at least.

plajjan:
BTW, try to include the command you are running when opening an issue :)

plajjan:
Closing this as I believe the issue is fixed with the merge of #842. @hallinde please check and reopen #841 if you can still see the issue.

PR #842, opened 2015-09-21
cli: force UTF-8 output on UTF-8 terminals

### LINKED ISSUE #841: raw_input is not UTF-8 safe
For some reason the print statements used by nipap CLI will happily print Unicode characters but the raw_input function will fail, such as https://github.com/SpriteLink/NIPAP/blob/master/nipap-cli/nipap_cli/nipap_cli.py#L1269

plajjan:
This is really rather weird but raw_input has an issue with UTF-8 output
even when stdout encoding is already set to UTF-8. Forcing UTF-8 for
UTF-8 terminals, which really should be a no-op, fixes that issue.

Since we are checking for UTF-8 output, this should be compatible with
legacy terminals.

Redirected or piped output is forced to UTF-8 just as before.

Fixes #841.

### DIFF
--- a/nipap-cli/nipap
+++ b/nipap-cli/nipap
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 """ NIPAP shell command
 
     A shell command to interact with NIPAP.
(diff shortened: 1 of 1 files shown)


garberg:
Weird, but looks good to me :D

PR #846, opened 2015-09-22
nipapd: fix 0 update

### LINKED ISSUE #843: pynipap.NipapError: <type 'exceptions.UnboundLocalError'>:local variable 'q' referenced before assignment
I believe I've found a bug - I'm not entirely sure how I'm triggering it, and what I'm doing is likely wrong, but I don't think it should be returning this response

I've tracked down the erroring code to https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap/backend.py#L991

plajjan:
We would get an exception when 0 objects were updated by a query due to
a bug in this code which is now being handled.

Fixes #843.

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -987,16 +987,19 @@ class Nipap:
                 }
             )
 
-        # We can have zero modified rows. Deal with it.
-        if len(qps) > 0:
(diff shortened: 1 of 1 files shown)


garberg:
Nice. Merging!

PR #848, opened 2015-09-23
nipapd: enable search on prefix-length

### LINKED ISSUE #847: Allow me to search on prefix_length
I wanted to search for all allocations that are larger than a /32 (ipv4) or a /128 (ipv6), so I cooked the following PHP-code:

```php
$result = $client->search_prefix(array(
    'auth' => $auth,
    'query' => 
        array(
            'operator' => 'and',
            'val1' => array(
                'operator' => 'equals',
                'val1' => 'type',
                'val2' => 'assignment'),
            'val2' => array(
                'operator' => 'greater',
                'val1' => 'prefix_length',
                'val2' => '32')
        )
));
```

It seems that's it's not possible to search on prefix_length (which is kinda odd, IMHO :)). Plz implement this. Thanks!

plajjan:
This enables search based on prefix-length as a match. It's using the
masklen() function in Postgres to do this and there is currently no
functional index on that so it might be slow.

Fixes #847.

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -400,6 +400,10 @@ _prefix_spec = {
             'column': 'inp.prefix',
             'ro': False,
         },
+        'prefix_length': {
+            'column': 'masklen(inp.prefix)',
+            'ro': True,
+        },
         'status': {
             'column': 'inp.status',
             'ro': False,


garberg:
Oh, that's a nice one! I thought we already had that implemented, but apparantly not :smile:

Issue #851, opened 2015-09-29
Error in webpage instructions


hallinde:
On page http://spritelink.github.io/NIPAP/docs/config-www.html
it says 
nipap-passwd add --username myuser --password mypassword --name "my user"

However I got this, 
someone@somewhere:/home/someone# nipap-passwd add --username jim --password jim --name "Jim"
usage: nipap-passwd [-h] [-u USER] [-p PASSWORD] [-n NAME] [-t] [-r]
                    [-f DB_FILE] [-c CONFIG] [--version]
                    [{list, add, delete, test-user, latest-version, create-database, upgrade-database}]
nipap-passwd: error: unrecognized arguments: --username jim

plajjan:
Fixed through merge of #852. It is now consistent.

PR #867, opened 2015-12-08
changed line 8 in config

genokan:
still had --username  and did not work on my install 
see #851 and #852


plajjan:
I'm not going to merge this as --username is what is in git master. I understand that it can be confusing since the latest released version (0.28.4) uses --user.

It's been a while since we made a release, mostly due to my and @garberg current situation at work / home, but I'll put in some effort into this :)

PR #872, opened 2016-01-04
Add option to modify existing users with nipap-passwd

plajjan:


nipap/nipap-passwd.man.rst
```diff
@@ -29,6 +34,10 @@ Options
 	delete
     	Delete user with username specifies with **-u** from the database
 
+    modify
```
garberg: There seems to be a mixture of tabs & space for indentation here. I guess space is what we want and thus the added entry is correct, but perhaps you could fix the other entries in the list of arguments while we're at it..?

garberg:
Looks good to me, just added a note regarding the indentation being inconsistent in the manual file. The branch is out-of-date so I cannot merge, fix that (and the indentation if you wish, otherwise we can take care of that some other time) and I'll merge!

plajjan:
Ok, fixed indentation and some other small thingy in the man page (was missing *).

garberg:
Nice, merging!

PR #890, opened 2016-02-17
Some IPv6 prefixes can't be looked up via the whois server

Eising:
The regex that matches VRF RT's also captures IPv6 addresses without letters in the first two sections of the address.

This pull request should fix that.

### DIFF
--- a/whoisd/nipap-whoisd
+++ b/whoisd/nipap-whoisd
@@ -99,6 +99,13 @@ class WhoisServer(SocketServer.StreamRequestHandler):
                 for p in res['result']:
                     response += format_prefix(p)
                     response += '\n'
+            # 1234:123:: - IPv6
+            res = Prefix.smart_search(query, { })
+            if len(res['result']) > 0:
+                total_hits += len(res['result'])
+                for p in res['result']:
+                    response += format_prefix(p)
+                    response += '\n'
         else:
             # respond with VRFs
             res = VRF.smart_search(query, { })


plajjan:
Explains why it worked for me (and I didn't notice this bug) - the Tele2 prefix is 2a00:800::/25 ;)

I agree on the bug but not necessarily the solution. I want to check that something is an IP through use of the socket module, or actually, looking at our current code - https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap/backend.py#L681 - we should probably stick to using IPy. Means we can't parse IPv4 addresses in the form 123.123123 (yeah, missing dot) but I think that's ok. In fact we can just copy the _is_ipv4, _is_ipv6 and _get_afi methods from the backend. I think it's actually cleaner copying them than importing the whole backend module.

This would allow us to write:
```
rt_match = re.match('^([0-9]+:[0-9]+):(.*)$', query)
if query == '':
    pass
elif rt_match and _get_afi(rt_match.group(2)):
    ....
```

I think this would work both for IPv[46] addresses and prefixes.

plajjan:
I'm not going to merge this because I don't think it's the right solution as I've outlined above. I do however agree that there is a problem here and I started looking into it, thus far leading to whoisd foreground to help debugging (#951), fix of VRF RT parsing in the backend (#952) and I found an an issue with IPv6 prefix/address parsing in the backend (#954). I'll try to get #954 fixed and then I'll tackle this one.

plajjan:
Opened #955 to track the problem. Closing this one.

PR #896, opened 2016-03-02
drop postgres dep

plajjan:
This removes the dependency on postgres from the nipapd package, allowing postgres to be installed on another machine. In addition nipapd can now be installed on the latest versions of Ubuntu as they have a newer postgresql version than we specified in our control file.


garberg:
Looks good, but it's out-of-date with the base branch. U plz fix?

PR #912, opened 2016-05-10
Handle extra_query in smart_search_pool and smart_search_asn

garberg:
Fix of two bugs in how extra_query was handled:
* ```smart_search_pool``` - handle queries without extra_query set
* ```smart_search_asn``` - pass extra_query to internal backend function


plajjan:
LGTM merging.

PR #918, opened 2016-05-14
Run jnipap test suite

garberg:
Despite the jnipap test suite being pretty small so far, we might just as well run it. I'm a bit unsure of how to perform the installations and tests properly, but it seems to work ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
     # -- common operations to apt install and virtualenv -----------------------
     # install dependencies for building packages and build NIPAP debian packages
     - sudo apt-get update
-    - sudo apt-get install -qq -y --force-yes devscripts python-docutils debhelper dh-python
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -58,6 +58,7 @@ install:
     # -- virtualenv build ------------------------------------------------------
     # install nipap dependencies
     - if [ "$INSTALL" == "venv" ]; then pip install -r nipap/requirements.txt; fi
+    - if [ "$INSTALL" == "venv" ]; then sudo apt-get install junit4; fi
```
plajjan: You need "-y" to avoid apt-get from prompting to install.

garberg:
Doh. Now I realized that there is an ```apt-get install``` in the file for dependencies common to the apt and virtualenv installations. Moved the installation of the junit4 packet there instead!

plajjan:
Nice :)

PR #929, opened 2016-06-15
fix: NameError: name 'options' is not defined

tiago-cruz-movile:
```
[root@nipap nipap-www]# nipap-passwd test-user -u cli -p cli
  File "/sbin/nipap-passwd", line 71, in <module>
    auth = af.get_auth(options.username, options.password, "nipap", {})
NameError: name 'options' is not defined
```


plajjan:
Nice, thanks!

Coincidentally I just noticed and fixed this now but I'll merge in your change instead :)

Issue #937, opened 2016-07-01
XSS vulnerability in tags


garberg:
The NIPAP web UI is vulnerable to an XSS attack where elements can be inserted into the DOM by adding tags with HTML tags to VRFs, pools or prefixes. Make sure tags names are escaped properly before writing them to the DOM.

Issue #887, opened 2016-02-15
Tooltips doesn't work on nipap-www


Eising:
On the latest release of nipap-www there are no tooltips anymore. As I'm not a long-time user, I can't pinpoint where it went wrong, and I can't help you with a fix either.

I've tried in several different browsers (chrome, safari and firefox), and none of them show anything when hovering over a comment icon, or mistyping a query. No errors in the javascript console either.

PR #938, opened 2016-07-01
Upgrade AngularJS and UI-Bootstrap

garberg:
Upgrade AngularJS and UI-Bootstrap to reasonable recent versions.
* AngularJS 1.5.3 (not the most recent version, but what UI-Bootstrap has been tested with)
* UI-Bootstrap 1.3.3

The original reason behind this was to get tooltips working again (see #887), but I also fixed the comment popups in the prefix list. In addition to this I did some cleanup among the JavaScript libraries and made sure we use minified versions of the libraries.

To get tooltips with HTML working with the new versions I had to disable the Strict Contextual Escaping-feature of AngularJS, see 79b5912 for more details. As a consequence to this it became apparent that NIPAP is vulnerable to an XSS attack where a user could insert elements into the DOM by crafting tags with HTML in them.I've greated #937 to handle this issue.


plajjan:
I can't review this code. Can only take a look at the result, i.e. look at the UI in a web browser. An idea that I just got around this.. wouldn't it be sweet if we could build docker container for every PR and could ask the nipap-demo machine to spin up a set of containers so it's easy to try out... or maybe I'm thinking too much about something that isn't *that* hard to try out today. Docker container images would be sweet regardless...

garberg:
I agree, this is pretty difficult to review. Spinning up a docker container so the web interface can be manually tested easily is an excellent idea. Sure, it's not very difficult to fetch the PR branch and perform tests locally, but automating it would lower the barrier.

We should also look into introducing unit tests for the parts of the web interface which can be tested. After all that's something which is put forward as a strength with AngularJS.

plajjan:
So it seems our psutil changes have made nipapd incompatible with Ubuntu 14.04 LTS. I think this is fine since 16.04 LTS is out but since my dev laptop is 14.04 it poses a slight challenge to me. Need to upgrade or build containers...

plajjan:
You know how trying to fix one problem leads to finding another problem and in trying to fix that problem you.. yeah, you get it.

I wanted to run this. I couldn't on my current dev env. So I wanted to put it in a docker container, which would be really sweet in lots of ways.. but to do that I wanted to fix a few other things, like #947. I think after #947 I can post some Dockerfiles so that I will be able to run everything in docker and thus I can test this :)

plajjan:
Yay @houndci-bot doing it's best not to be very helpful at all. Lemme see if I can docker the shit out of this PR =)

plajjan:
Neither the prefix comment or any other tooltips (I can't really remember where they are) are working for me :/ Tested using Chrome on Mac OS X with nipap-www served through docker.

plajjan:
My bad yo. Though I'd be clever and tag the image I built for this and then I failed to use the docker compose file I had updated with the tag.

Now that I really tried it out: This thing is working beautifully! LGTM et al. Will merge.

PR #943, opened 2016-07-28
Support unlimited max_result for search_prefix

### LINKED ISSUE #466: Some prefixes are grey
I would like to bring this issue back. I already described it here https://github.com/SpriteLink/NIPAP/issues/405#issuecomment-24314025

I was advised by plajjan that this issue should be a problem with the system, not with the database.
So I set up a fresh new ubuntu, installed nipap according to the 'Getting started' guide and imported my database using my import script (https://github.com/dosomder/NIPAP/commit/2b109c4cd27ce5218d006a99edab7978798e6d6c).
```
root@ubuntu:/tmp# nipapd --version
nipapd version: 0.23.2
```

However, the issue is still present:
problem

I hope we can solve this. Thanks

plajjan:
For certain queries, like when expanding a prefix by clicking the '+' in
the web UI we really shouldn't have a limit. Unlimited is signified by
setting max_result to False or None.

Downside might be DoS attack on the backend or some incorrect logic
somewhere blowing it up with heavy queries inadvertently.

Fixes #466.

### DIFF
--- a/nipap-www/nipapwww/public/nipap.js
+++ b/nipap-www/nipapwww/public/nipap.js
@@ -1768,7 +1768,7 @@ function collapseClick(id) {
 		search_q.include_children = false;
 		search_q.children_depth = 1;
 		search_q.parents_depth = 0;
-		search_q.max_result = 1000;
+		search_q.max_result = false;
 		search_q.offset = 0;
 		delete search_q.indent;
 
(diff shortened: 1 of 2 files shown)


plajjan:
This is written blind, i.e. I have not tested it at all since my Linux dev laptop is currently under maintenance.

@Eising, could you try and see if this fixes it for you? :)

PR #947, opened 2016-08-02
Revamp database schema install & upgrade

plajjan:
This moves the db schema install & upgrade functionality from the debian install scripts into nipapd itself which makes it more portable (guaranteed to run everywhere nipapd runs, duh).

I've changed the debian install scripts accordingly. For example, the very initial setup, like creating a db user and the database itself still happens in the debian postinst scripts as it requires super-user privs.

To make it easy to access the schema data I copied the sql/* files into nipap/db_schema.py -  simply as long strings. This makes it considerably more difficult to read and write. The .plsql files look beautiful in my vim but are rather plain as strings in a python script. I could of course revert this and just read the .sql files instead.

Opinions?

There's quite a few different scenarios in these scripts. I have made some testing but not sure I've tested it all. Testing is highly appreciated!


plajjan:
Actually, just writing about this has me leaning towards not putting the sql schema in a python file. But the problem is that I don't know where to find the .sql files. If we installed them with debian they'll be in /usr/share/nipap/sql but if a developer is running this they are somewhere else. What to do?

Another approach is to have the Makefile take the .sql files and put them in a .py file... makes it easy to read/write yet easy to work with from python... but make has to run first, confusing? Maybe.

plajjan:
I should probably test the db schema upgrade too. I haven't really done that in detail, i.e. defining new upgrade schema and bumping db version number.

nipap/nipap/backend.py
```diff
@@ -758,17 +763,53 @@ def _connect_db(self):
...
-            self._con_pg.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
-            self._curs_pg = self._con_pg.cursor(cursor_factory=psycopg2.extras.DictCursor)
-            self._register_inet()
-            psycopg2.extras.register_hstore(self._con_pg, globally=True, unicode=True)
-        except psycopg2.Error as exc:
-            self._logger.error("pgsql: %s" % exc)
-            raise NipapError("Backend unable to connect to database")
-        except psycopg2.Warning as warn:
-            self._logger.warning('pgsql: %s' % warn)
+        while True:
```
garberg: ```while True:``` without any kind of throttling always rings some warning bells for me, but I guess we're good here :smile:

garberg:
Many changes and my knowledge of the apt and Travis stuff is somewhat limited, but things looks good from what I can see.

Should we ship a pre-built db_schema.py? It's a bit redundant, but I guess the convenience of being able to run the code directly from the git repo (and not needing to remember to rebuild the file when pulling new versions...) is more important.

Rebase on master and I'll merge!

nipap/nipap/backend.py
plajjan: I agree on the concern but don't see what we could do differently.. more error handling if anything but I don't really know how or for what. Let's see what error conditions we might run into.

plajjan:
I also had concern about merging this now, which will go into 0.29.1 or so but.. I think we'll go. Ut med båten och se om den flyter!

Since you've given go-ahead I'll push the merge button =)

PR #950, opened 2016-08-03
cli: read connection params from env or prompt

### LINKED ISSUE #838: Read password from stdin instead of .nipaprc
Hello together

Even if the webinterface is really great I'm kinda the CLI-Guy. I tested it and like it - besides that part about storing your password clear text in an file. 

If you're using the LDAP-Backend you're most likeley having several systems you have access to. And you don't want to give the server admin(s) or some botnet the opportunity to steal your password and gain more systems.

Therefore I'd like to add a feature request: Instead of exiting with an error if "password" is not found at .nipaprc give the user a prompt (e.g. using "getpass") and only raise the exception if thats also not provided.

plajjan:
In addition to reading the connection parameters from the .nipaprc
configuration file we now also try to get them from environment
variables. We prefer environment variables, then from .nipaprc and last,
in the case of the password, we will interactively prompt for it.

nipapd host and port now also have the defaults 'localhost' and 1337
respectively.

Fixes #838.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -46,17 +46,28 @@ def setup_connection():
     """ Set up the global pynipap connection object
     """
 
-    # build XML-RPC URI
+    # get connection parameters, first from environment variables if they are
+    # defined, otherwise from .nipaprc
     try:
-        pynipap.xmlrpc_uri = "http://%(username)s:%(password)s@%(hostname)s:%(port)s" % {
-                'username': cfg.get('global', 'username'),
-                'password': cfg.get('global', 'password'),
-                'hostname': cfg.get('global', 'hostname'),
-                'port'    : cfg.get('global', 'port')
-            }
-    except (ConfigParser.NoOptionError, ConfigParser.NoSectionError):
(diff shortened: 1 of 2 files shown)


plajjan:
Reading from env vars also makes it easier to put this in a docker container :)

nipap-cli/nipap_cli/nipap_cli.py
```diff
@@ -46,18 +46,29 @@ def setup_connection():
...
+    if con_params['password'] is None:
+        import getpass
+        con_params['password'] = getpass.getpass()
+
+    # build XML-RPC URI
+    pynipap.xmlrpc_uri = "http://%(username)s:%(password)s@%(hostname)s:%(port)s" % con_params
+
     ao = pynipap.AuthOptions({
         'authoritative_source': 'nipap',
         'username': os.getenv('NIPAP_IMPERSONATE_USERNAME') or cfg.get('global', 'username'),
```
garberg: ```os.getenv('NIPAP_USERNAME')``` should be added here as well, right?

plajjan: Yeah, you're right. I use con_params['username'] in my latest patch.

PR #953, opened 2016-08-03
Fix broken parsing of VRF RT

plajjan:
The backend smart parsing function currently builds broken queries for stuff that looks like a VRF RT, like '123:456'. This fixes that and adds a test case to make sure we catch this in the future.


garberg:
Looks good :+1:

PR #956, opened 2016-08-04
backend: fix smart parsing of IPv6 prefixes

### LINKED ISSUE #954: IPv6 addresses parsed as VRF RT
The new smart parser isn't so smart after all - it misinterprets IPv6 addresses like 2001:1000::/32 as a VRF RT :(

plajjan:
The new smart parser that I wrote didn't correctly parse IPv6 addresses,
instead mistaking them for VRF RTs, which are also colon separated
numbers. We didn't have a test case for this and I believe I didn't
catch this earlier as I probably use the Tele2 IPv6 prefix 2a00:db8::/25
that contains hexadecimal characters, thus differing from a VRF RT.

I have now added the explicit classes ipv6_address and ipv6_prefix which
also lead to their own expansion from AST to dictsql.

There are now also test cases to cover shortened IPv4 prefixes (like
10.0/8) and IPv4 prefixes with bits set in host part so we check that
the strict prefix mangling works.

Fixes #954.

### DIFF
--- a/nipap/nipap/smart_parsing.py
+++ b/nipap/nipap/smart_parsing.py
@@ -102,6 +102,10 @@ class SmartParser:
         comp_ipv4_address = Combine(ipv4_oct + ('.' + ipv4_oct*3))
         ipv4_address = Combine(ipv4_oct + ('.' + ipv4_oct*3)).setResultsName('ipv4_address')
 
+        # IPv6 address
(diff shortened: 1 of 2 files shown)


plajjan:
So, in the AST there's a different type/class for ipv6_address and ipv6_prefix but there's no similar ipv4_address or ipv4_prefix. The reason for this is that IPv4 can be shortened, like we can search for 10/8 and the way we treat these is that we check with IPy if it can be correctly interpreted as an IPv4 prefix and only then do we treat it as such, otherwise we fall back to interpreting it as text. Since pyparsing only uses regexp I can't really (or would be hellish long regex) check it in there. IPv6 on the other hand doesn't have such a shorthand format and so is actually easier to parse in pyparsing, although that also results in a hellish long regex.

I realise as I write this that I could have used a separate class for ipv4_address since I think we check that we have four octets, i.e. while we can expand 10/8 into 10.0.0.0/8 we would never try to expand 10 into 10.0.0.0. Anyway, net effect is the same so I'll not do that right now.

nipap/nipap/smart_parsing.py
```diff
@@ -429,8 +433,6 @@ def _string_to_dictsql(self, part):
                     'val2': part[0]['word']
                     }
 
-        elif part.getName() == 'ipv4_address':
-            dictsql = {}
```
garberg: I'm not quite sure how the other changes introduced relates to this change - was this never used?

garberg:
Except from me not understanding the parsing in detail, it looks good. Merge!

nipap/nipap/smart_parsing.py
plajjan: No, it was never used. This is what I tried to describe above.

There are three formats involved here
 * text query
 * AST
 * dictsql

pyparsing parses a text query and outputs the AST (abstract syntax tree). Each element has a type, like 'ipv6_prefix'. The part you are looking at above is in the code that turns the AST into dictsql. For IPv4 we don't have a separate type but actually classify it as text / "else" and the classification of IPv4 or truly text is done in the ast to dictsql part.. thus this would never match and so I removed it.

plajjan: There is actually an ipv4_address matching defined at https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap/smart_parsing.py#L103

but it's not in use. You can see how all other things are brought together in the parsing tree here https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap/smart_parsing.py#L117-L132 and that's where this PR adds ipv6_address & ipv6_prefix but there is no ipv4_addres. Things that are IPv4 addresses will be matched by the 'word' type.

PR #959, opened 2016-08-04
backend: switch nipap.conf.dist to Jinja2 syntax

### LINKED ISSUE #949: Use Jinja2 to produce initial config
We currently use sed in the debian postinst script to rewrite certain things in the nipap.conf.dist and write nipap.conf.

Looking at writing a Dockerfile I reckon it would be useful to write nipap.conf using Jinja2 instead of dumb sed commands.

How about we rewrite nipap.conf.dist to be a Jinja2 template and use Jinja2 to populate values. It will still be easy for a human to change if that's the case.

plajjan:
I've switched out some of the values that we want to substitue to Jinja2
style variables so that it's easier to use, well, Jinja2 to do the
rendering of the config file.

This is kind of preparatory work for putting nipapd in a docker
container as I envision using envtpl to do the variable subsitution.

The debian nipapd postinst script is however still using sed since I
didn't want to include envtpl in the normal debian package and I'm too
lazy to write a Jinja2 renderer right now. This should still be better
than before IMHO.

Fixes #949.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,14 +61,14 @@ install:
     # SQL
     - if [ "$INSTALL" == "venv" ]; then sudo su -c "cd nipap/sql; make install" postgres; fi
     # move configuration file into place
-    - if [ "$INSTALL" == "venv" ]; then sudo mkdir /etc/nipap; sudo cp nipap/nipap.conf.dist /etc/nipap/nipap.conf; fi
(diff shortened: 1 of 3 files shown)


garberg:
A nice change!

However, unless the user runs NIPAP in a Debian derivate or a Docker container the PR changes from a perfectly fine nipapd.conf.dist which more or less can be used out-of-the-box to something which will crash and burn :wink: I'll merge this now and create a separate issue for creating a tool which renders the nipap.conf.dist-file, which also can be used from setup.py.

plajjan:
Sure, the old config was almost usable out of the box.. but since the user has to create a postgres account I don't think it's much to ask to replace a few things. Not like it's difficult to guess what goes where ;)

PR #960, opened 2016-08-04
Dockerify nipapd

plajjan:
This adds a Dockerfile for nipapd and an entrypoint script which will
write the config file and launch nipapd.

This depends on #959.


garberg:
LGTM. Merge when rebased!

Issue #974, opened 2016-08-09
CLI breaks with non-ASCII characters


garberg:
A new problem with non-ASCII characters in the CLI has been introduced:

```
$ nipap a a prefix 10.0.0.0/24 type reservation description åäö
Network 10.0.0.0/24 added to VRF 'default' [RT: -]: åäö

$ nipap a l 10.0.0.0/24
Searching for prefixes in any VRF...
Traceback (most recent call last):
  File "/usr/bin/nipap", line 131, in <module>
    cmd.exe(cmd.arg, cmd.exe_options, args)
  File "/usr/lib/python2.7/dist-packages/nipap_cli/nipap_cli.py", line 489, in list_prefix
    len(str(val)))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```

garberg:
OK, added a pull request for this. Two additional considerations:
* As this isn't the first time we're breaking non-ASCII characters we should probably look into a way to test this. Perhaps BATS (https://github.com/sstephenson/bats - available in main Debian & Ubuntu repos) might be useful?
* Perhaps we should consider letting the NIPAP CLI be Python 3 only to get rid of this shite once and for all.

plajjan:
Fixed by #976. Closing.

PR #975, opened 2016-08-09
Handle non-ASCII characters in the CLI

garberg:
Some fixes to make the CLI handle non-(7-bit)ASCII characters again. I chose to simply replace all instances of str() with unicode() to keep things simple, even at locations which could not possibly return non-ASCII characters such as prefix lengths.

### DIFF
--- a/nipap-cli/nipap_cli/nipap_cli.py
+++ b/nipap-cli/nipap_cli/nipap_cli.py
@@ -507,7 +507,7 @@ def list_prefix(arg, opts, shell_opts):
                 try:
                     col_def['pool_name']['width'] = max(col_def['pool_name'].get('width', 8),
                                                         len(unicode(p.pool.name)))
-                except:
+                except AttributeError:
                     pass
             # override certain column widths
             col_def['type']['width'] = 1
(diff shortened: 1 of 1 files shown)


plajjan:
So the one thing I have against this is that unicode() doesn't exist in python3. I would much rather move in the direction of always running the CLI with python3. I think py3 exists on all platforms today so it shouldn't really be an issue. Are all the libs we use py3 compatible?

I think the py2&py3 safe way is to do .decode('utf-8') on the input, can you try?

plajjan:
Or we could do some hack, like 
```
import sys
if sys.version_info[0] == 2:
    str = unicode
```

garberg:
Hehe, yep, I wrote something similar in #974 :smile:

I agree, aiming towards Python 3 is the better way forward. There are some other things (exception handling, print statements, ...) which needs to be addressed for full Python 3 compatibility, which is why I chose this easier way forward for now.

I'll do some decode('utf-8'):ing, hang on...

plajjan:
Ok, I get that but I really don't want to take a step in the wrong direction, away from py3 compatibility because then we only adds things to the list of things to do and that's.. well, it's no good.

import the new print statements and change exc handling as well? :)

Does the CLI rely on some py2-only lib?

plajjan:
Not relevant anymore after merge of #976 - we now do full py3 compatible instead of fiddling with .decode() & .encode() in py2.

PR #976, opened 2016-08-09
Make CLI client Python 3 only

garberg:
A first try to make the NIPAP CLI client Python 3 only, hopefully avoiding more unicode errors in the future. Consider this a work in progress!

At https://wiki.python.org/moin/PortingToPy3k/BilingualQuickRef#codecs it's mentioned that the following could be done to safely write unicode to STDOUT with LC_ALL=C:
```
sys.stdout = io.TextIOWrapper(sys.stdout.detach(), encoding='UTF-8', line_buffering=True)
```
I'm a bit hesitant to whether we should add it. Not adding it causes this to be written when trying to write non-ASCII characters with LC_ALL=C, as the C locale is considered ASCII only:
```
Crazy encoding for prefix 10.0.0.0/24
```

From what I remember an annoyance was that you'd get unicode errors when redirecting output containing non-ASCII characters. I've not seen this behaviour in the Python 3 version.

I also made some changes to the Debian packaging config (incorporating recommendations from https://wiki.debian.org/Python/LibraryStyleGuide) which probably can be refined a bit. Many warnings are displayed the package is built.


nipap-cli/nipap
```diff
@@ -15,20 +15,6 @@ import nipap_cli.nipap_cli
...
-    # print statements but raw_input() will fail when the prompt contains
-    # Unicode characters. Forcing the output to UTF-8 fixes that, despite
-    # setting UTF-8 output when output encoding already is UTF-8 should be a
-    # no-op. Note how we only do this when the encoding is already set to UTF-8
-    # so any legacy terminals should still work fine.
-    if sys.stdout.encoding == 'UTF-8':
-        sys.stdout = codecs.getwriter('utf8')(sys.stdout)
-else: # we are likely being redirected / piped to something else
-    # force UTF-8 output for redirected stdout
-    sys.stdout = codecs.getwriter('utf8')(sys.stdout)
```
plajjan: Have you or can you verify that UTF-8 actually gets piped correctly to the next process?

nipap-cli/nipap_cli/command.py
```diff
@@ -194,10 +191,10 @@ def parse_cmd(self, tree, inp_cmd = None):
             if self.children is not None:
                 self.key_complete = False
                 match = False
-                for param, content in self.children.items():
+                for param, content in list(self.children.items()):
```
plajjan: Hmm, isn't this a dict? Why would 2to3 put a list around that? Doesn't make sense IMHO and I don't understand how it can work.

plajjan:
Most of the changes are to update to print() instead of old style print.. I can't say I've looked at each and every one in detail but it looks good overall after a cursory look.

The iteration over a dict with key and value using .items() seems to now be wrapped in list(), after the 2to3 tool had its way, which I don't understand how it could work.

Travis-CI is failing because nipap_cli is imported from nipaptest.py and is thus run with Python2. The actual issue is that print statement that doesn't conform to py2. Can this be fixed through `from __future__ import print_statement`?? We can potentially run the whole tests using py3.. they are currently executed with "nosetests ....py" and nosetests will be run with whatever is specified by the Travis-CI build matrix but I think we can override that.

garberg:
In Python 3 the .items()-method no longer returns a list. I guess 2to3 wraps all uses of .items() in a list() to be sure that it doesn't introduce some bug due to this. In our case I'd say we can remove the list()-functions. I fix.

I'm not sure ```from __future__ import print_statement``` will solve the testing issue, and even if it does I guess we need to ask ourselves if we want to keep the CLI code both Python 2 and Python 3-compatible. When I initiated the transition of the CLI from Python 2 to Python 3 I worked under the assumption that we'd go for Python 3 only, but at that time I wasn't aware that the tests imported both CLI and backend modules.

nipap-cli/nipap
garberg: Not more than running ```nipap a l | tee``` and ```nipap a l > foo``` with some non-ASCII characters in prefix descriptions and verifying that I can see them in the resulting output.

plajjan: Ok, good enough :)

plajjan:
Ok, list() seems redundant, please remove. I think it's cleaner looking without. http://stackoverflow.com/questions/17695456/why-python-3-needs-wrap-dict-items-with-list


No, I don't really aim for py2 compatibility of the CLI but simply adding that import will probably make the tests pass. nipaptest.py is big, perhaps too big. In addition to importing and testing the CLI it also imports backend stuff, which we know is NOT py3 compatible. So unless you want to take on fixing the backend too, we would need to try and make CLI py2 compatible. If it's as simple as the import statement then let's just do it. If it requires more things then no, fuck it. I think there's a distinction between hacking it to run py2 for CI and making it py2 compatible for all users (more utf-8 hackery n stuff).

If import statement doesn't fix it then I think we need to split up nipaptest.py into smaller parts - that might be a good idea regardless.

nipap-cli/nipap_cli/command.py
```diff
@@ -228,7 +228,7 @@ def parse_cmd(self, tree, inp_cmd = None):
             # len(p) == 0
 
             if len(p) != 0 and len(self.key) == 1:
-                key, val = self.key.items()[0]
+                key, val = list(self.key.items())[0]
```
garberg: Here we need to keep the list(), as we're selecting an element from the list.

garberg:
OK, after a bit of wrestling with Travis-CI the tests are run and passes again.

As the python environment causes problems when part of the project is Python 2 and part of the project is Python 3 I chose to avoid the python-specific build environments and use a generic image instead. This, naturally, introduced some new problems but I think I've been able to address them all.

Two issues regarding postgresql authentication turned up, and I honestly don't understand why it bacame an issue now, or rather, why it did work before.
* When doing the venv-installation, I needed to explicitly set the PGPASSWORD environmental variable to the hard-coded password used during the tests.
* When doing the apt-installation, the db_host config variable was set to "" by the postinst script when the database host provided to debconf is "localhost". This causes postgresql to use peer authentication, and as the tests are not run as the user ```nipap``` this failed. To address this, I chose to simply rewrite the db_host variable in nipap.conf to localhost in the Travis-CI config, befort executing the tests.

Also, as the Travis image runs Ubuntu Trusty, which is shipped with an ancient version of psutil, I had to implement a work-around to support both old (pre 2.0.0) and new versions.

plajjan:
LGTM

PR #980, opened 2016-08-11
Handle unclosed single quote

garberg:
Handle unclosed single in addition to double quotes, including test cases. Also added test cases for unclosed parentheses.


plajjan:
LGTM, merging.

PR #983, opened 2016-08-11
Correct response on smart parser error

### LINKED ISSUE #982: Response on smart parser error incorrect
The response sent when the smart parser encounters an error does not follow the same format as for a successful query, which leads to parser error messages not being displayed:
```
$ ./nipap --show-interpretation a l 'foo "'
Query interpretation:
No addresses matching 'foo "' found.
```
Here I'd expect the error message to be displayed.

garberg:
Update the format of the result returned when the smart parser encounters a ValueError to match the usual result dict. I chose to add some empty values as well, to simplify client implementation.

Result from the CLI client:
```
$ ./nipap --show-interpretation a l 'foo "'
Query interpretation:
       `-- foo ": unclosed quote, please close quote!
   This is not a proper search term as it contains en uneven amount of quotes.
No addresses matching 'foo "' found.
```

Fixes #982

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -1664,13 +1664,17 @@ class Nipap:
             query = self._parse_vrf_query(query_str)
         except NipapValueError as exc:
             return {
-                'interpretation': [
-                    {
+                'interpretation': {
+                    'operator': None,
+                    'val1': None,
+                    'val2': None,
+                    'interpretation': {
                         'string': query_str,
-                        'interpretation': exc,
-                        'attribute': 'text'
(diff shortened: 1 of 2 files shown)


plajjan:
I realise this code is very naive. In the old parser I think unclosed quote was pretty much the only thing we caught. With the new parser we also have unclosed parentheses which I suppose we won't be able to distinguish now, right?

plajjan:
@garberg Still, this takes us part of the way so I'm ready to merge this although I think we should catch more specific exceptions. Do you wanna have a go at that in this PR or should I merge and we leave that for later?

PR #989, opened 2016-08-22
Merge nso-nipap

garberg:
Add a Cisco NSO/former tail-f NCS NIPAP client library.


plajjan:
Not tested but the whole nso-nipap thing is rather experimental so I'll merge this anyway. Better to let people see it exists than hide it away.

PR #991, opened 2016-08-24
pynipap.py commented documentation additions

crosson:
Updated the pynipap.xmlrpc_uri variable in the example given to match the default port and added the required corresponding path, "/XMLRPC".

Also added a note remote API connections.

Added an example on searching.


plajjan:
Looks good. Thanks for your contribution! :)

crosson:
Ah shucks. Looks like Sphinx didn't parse my changes correctly. They aren't wrapped into neat text boxes and some of the additions are doubled up on the same line. Figures!

PR #993, opened 2016-08-24
Correcting Sphinx Markup in pynipap.py

crosson:
My previously added notes, in PR #991, were missing some necessary markup causing
sphinx to render the notes incorrectly. Correcting this.

I've tested this with a local sphinx-build and it appears correctly now.


plajjan:
Cool!

PR #996, opened 2016-08-30
Continue on error

garberg:
Continue with the next request when an error occurs in the allocation of prefixes. Before this change, NSO would lock up when for example trying to allocate prefixes from a pool without free prefixes.


plajjan:
I haven't tested this. We're so early on in the life of this that there's probably plenty of bugs everywhere and every PR is an improvement ;) Alpha alpha. Code LGTM - will merge.

PR #1006, opened 2016-09-02
www: Handle top-level error flag from smart_search

### LINKED ISSUE #1004: Broken prefix list page web UI
Check out nipap-demo.spritelink.net - I'm just getting a big ugly error message: "undefined".

I ran nipapd with -d -f and it appears to work OK.

Could this be that the web-UI expects the interpretation result to look differently?

garberg:
The "top-level" error flag newly introduced in the smart_search_-functions collided with an error flag used by the web interface. This is a quick fix for that issue which simply removes the error key from the result received from smart_search. The flag is currently only used to signal query string parser errors, which also are present deeper in the interpretation dict where it will be picked up and displayed by the web UI. Thus no functionality is lost when the top-level flag is removed.

I was probably a bit too quick implementing the error flag; we should probably think through how to convey "soft errors" (such as string parsing errors) vs "hard errors" (such as, for example, the backend
being unable to reach the database) and how to do it in a future JSON-based REST API.

Fixes #1004.

### DIFF
--- a/nipap-www/nipapwww/controllers/xhr.py
+++ b/nipap-www/nipapwww/controllers/xhr.py
@@ -109,6 +109,12 @@ class XhrController(BaseController):
             result = VRF.smart_search(request.params['query_string'],
                 search_options, extra_query
                 )
+            # Remove error key in result from backend as it interferes with the
+            # error handling of the web interface.
+            # TODO: Reevaluate how to deal with different types of errors; soft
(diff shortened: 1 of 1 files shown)


plajjan:
I'm afraid we will forget about this. Should we at the very least put some comments in there to hint about what should be done in future rework of this code?

garberg:
The unspoken plan was to create an issue :smile:

Added comments.

PR #1009, opened 2016-09-09
Dockerify nipap-www

garberg:
Added a dockerfile and entrypoint-script for nipap-www. It's based on a standard ubuntu:xenial-image, serving the NIPAP web UI with apache2 with mod_wsgi.

All installation of dependencies and NIPAP packages is done in the dockerfile, while the local auth database is setup, permissions are corrected and apache2 configured in the entrypoint shell script. The
dockerfile currently doesn't provide any way to make changes to the nipap-www.ini-file.

### DIFF
--- a/Dockerfile.www
+++ b/Dockerfile.www
@@ -1,0 +1,61 @@
+# This file describes a docker image for running nipap-www in docker
+#
+# Build the docker image:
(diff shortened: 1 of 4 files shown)


plajjan:
LGTM

Issue #934, opened 2016-06-23
avps attributes can't contain semicolons


rfederici:
I receive no problems when pushing via pynipap, my avps attribute can contain things like {"hhd_info":"1TB RAID0; 2TB RAID0"}, and it's viewable on the web application.

But when I try to edit the script-created host via the webapp, I can't edit anything (description, other avps tags) without a 500-error popup. Had to go through a lot of trial and error to figure it out.

This bug is very minor, but if there's a reason why it's not sanitized, a relevant error message would be nice!

PR #1011, opened 2016-09-09
POST data to XHR functions

garberg:
Okay, another PR with way too many changes :smile:

To solve the problem with not being able to enter extra attributes with semicolons in them, I chose to change how the client side of the web interface sends data to the server side. Previously this was done with HTTP GET requests and the data was not URL encoded (thus the problem with semicolons, which actually wasn't limited to extra attributes, no fields could contain semicolons).

Instead of performing all this URL encoding and decoding, it seemed as a better solution to send the data (which was already in the form of a JavaScript object) as JSON in the body of a HTTP POST request, using a shorthand ```request.json``` attribute to decode the data into Python dicts at the receiving side. As the previous code sort of depended on behaviour specific to how the HTTP GET query strings are generated (for example omitting null values and passing everything as strings), the validation performed at the server side needed to be modified.

All in all this is also a step towards minimising the code base at the server side of the web application and letting the client side of the web interface speak directly to the API. One natural next step forward in this direction would be to move some of the code doing validation and determining what values to send to the backend from the server side XHR controllers to the client side AngularJS controllers. 

Lastly, while I was at it, I also changed how the ID of objects to remove is sent to the server to match the ```edit_```-functions. Instead of sending it as part of the query data (.../remove_foo?id=X), it's now part of the path (.../remove_foo/X).

### DIFF
--- a/nipap-www/nipapwww/controllers/xhr.py
+++ b/nipap-www/nipapwww/controllers/xhr.py
@@ -34,20 +34,20 @@ class XhrController(BaseController):
 
         # TODO: add more?
         attr = {}
-        if 'id' in request.params:
-            attr['id'] = int(request.params['id'])
-        if 'prefix' in request.params:
-            attr['prefix'] = request.params['prefix']
-        if 'pool' in request.params:
-            attr['pool_id'] = int(request.params['pool'])
-        if 'node' in request.params:
-            attr['node'] = request.params['node']
-        if 'type' in request.params:
-            attr['type'] = request.params['type']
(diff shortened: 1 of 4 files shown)


garberg:
Oh, and this fixes #934.

plajjan:
Built docker container, did a bunch of monkey user testing on it. Seems fine. I had a brief look at the code which also seems quite fine, although I really do dislike some of our structure but it's not your fault and as you pointed out we can improve further. This is good for now - merging!

PR #1012, opened 2016-09-09
Fix prefix smart search with tags

### LINKED ISSUE #1008: Prefix tag search doesn't work
The prefix tag search doesn't work any more:
```
$ nipap address list "#foo"
Command failed: <type 'exceptions.TypeError'>:string indices must be integers, not str
```
Searching for tags in pools and VRFs still works though.

garberg:
Fix prefix smart search where the search string contains tags and adjust debug logging statements to display the correct value.

Also adjusted how the search interpretation is displayed slightly.

Before:
```#foo: tag must contain #foo``` - a bit misleading as the tag name in this case actually is ```foo```, without hash.

After:
```#foo: tag must contain foo``` - correct as the tag name is ```foo```.

Fixes #1008.


plajjan:
LGTM

PR #1018, opened 2016-09-11
Improve handling of extra attributes with empty values

### LINKED ISSUE #882: Ignore extra attributes with empty name in web UI
The web UI should ignore Extra attributes that have an empty name so that they are not sent to the backend. As discussed in #876 maybe we should warn when an extra attribute with an empty name but a value is set. If however both the name and value is empty I think we should silently ignore it as the user has likely just pushed the "add attribute" button and then save.

Related to #876.

garberg:
In the web UI, improve the handling of extra attributes with empty values so that:
* Entries where both the attribute and value is empty are removed when sending data to the server.
* The attribute field is mandatory when a value is entered.

Fixes #882.


plajjan:
LGTM.

PR #1020, opened 2016-09-13
nipapd: mangle IPv4 address

### LINKED ISSUE #860: Invalid IP address causes exception
A search for 10.18.01.1 in the web UI will pop up an exception:
```
invalid IP4R value: "10.18.01.1" LINE .....
```

I guess the socket check is more forgiving than ip4r.

plajjan:
This mangles IPv4 address in search queries through IPy to get a
properly formatted IPv4 address before we send it to Postgres. We do
validation of IPv4 addresses but once they are verified we send the
original value, so for example 10.18.01.1 is recognized as IPv4 but
Postgres doesn't like the leading 0 in the third octet. By running it
through IPy we get rid of that leading 0 and all is fine.

Fixes #860.


garberg:
Looks nice, merging!

PR #1031, opened 2016-09-14
www: Add slash to generated URI

### LINKED ISSUE #1029: Opening VRF selector reloads page
Opening the VRF filter selector causes the web page to reload and thus the VRF selector to disappear.

garberg:
After the AngularJS upgrade from v1.3.15 to v1.5.3 a slash has been introduced in the URI from some reason. The code in NIPAP which stores the current search state in the URI neeed to be updated to reflect this, to avoid the page from reloading when opening the VRF selector.

Fixes #1029.


plajjan:
one-char-commit-fest!

PR #1037, opened 2016-09-17
Accept parent_prefix as None

### LINKED ISSUE #998: Prefix search_options returns Null
in pynipap using Prefix.search:

when doing a search query without specifying any search_option, the returning search_option is {'include_all_children': False, 'max_result': 50, 'include_neighbors': False, 'parent_prefix': None, 'include_all_parents': False, 'parents_depth': 0, 'offset': 0, 'children_depth': 0}

when trying to reuse this dict in a new search it raises Nonetype error (probably to the fact that parent_prefix is 'None':
    prefixes = Prefix.search(query, search_options)
  File "/usr/lib/python2.7/dist-packages/pynipap.py", line 936, in search
    raise _fault_to_exception(xml_fault)
pynipap.NipapError: <type 'exceptions.TypeError'>:int() argument must be a string or a number, not 'NoneType'

plajjan:
It won't do anything but at least we avoid throwing an exception. This
can be useful if someone wants to feed the search_options returned by a
query into another query.

Fixes #998.

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -3383,7 +3383,8 @@ class Nipap:
                     ''' 'offset'. Only integer values allowed.''')
 
         # parent_prefix
-        if 'parent_prefix' not in search_options:
+        if ('parent_prefix' not in search_options or
+                search_options['parent_prefix'] is None):
             search_options['parent_prefix'] = None
         else:
             try:


nipap/nipap/backend.py
```diff
@@ -3383,7 +3383,7 @@ def search_prefix(self, auth, query, search_options=None):
                     ''' 'offset'. Only integer values allowed.''')
 
         # parent_prefix
-        if 'parent_prefix' not in search_options:
+        if 'parent_prefix' not in search_options or search_options['parent_prefix'] is None:
```
plajjan: Blargh. OK OK Hound. I'll fix.

garberg:
LGTM :+1:

PR #1042, opened 2016-09-19
Fix unicode searches and a bunch of operators 

plajjan:
This revamps part of the smart parsing so that we are able to use Unicode characters when performing searches. In addition I've added a couple of operators that we were missing and tests for all of the above.


garberg:
Despite Sandor Clegane's disapproval I think this looks good. Merging!

Issue #1007, opened 2016-09-02
SQL error when matching vlan attribute with string


garberg:
Trying to match the prefix VLAN attribute to a string leads to an SQL error being shown to the user:
```
$ nipap a l vlan=foo
Searching for prefixes in any VRF...
Command failed: invalid input syntax for integer: "foo"
LINE 97:                         WHERE  inp.vlan = 'foo' 
                                                   ^
```
We should probably implement a "catch-all" try-catch clause which catches all SQL errors and returns a more user-friendly error message.

plajjan:
Not just VLAN.. it's any integer attribute, such as 'indent'.

plajjan:
But your suggested fix will catch the error too late.. for "type" we reasoned differently and put the check in smart_parsing, so before the SQL query is run whereas here you suggest to do it afterwards.

Naturally we can do both, i.e. it might be useful to catch this in case someone feeds us a dictSQL in the search_prefix() RPC (not smart_search_prefix()) but I think the correct fix for normal search (which is smart_search based) is to catch it in the smart parser.

plajjan:
Half of this is done, i.e. we catch exceptions and throw more friendly error messages to the user but the other half is about actually catching the exception before we run the SQL query so that we instead can return an error in the search interpretation.

I'm closing this one as I've opened #1067 to track that last bit.

PR #1046, opened 2016-09-19
nipapd: Catch invalid values for integers

plajjan:
This catches errors from the database when the provided value is invalid
for the given data type, in this case integer.

Part of #1007.


plajjan:
Closing since we merged #1050 to fix this error instead.

PR #1049, opened 2016-09-20
WIP - nso-nipap rewrite

fredsod:
This is almost a complete rewrite of the nso-nipap cdb subscriber.

Todo:
- [ ] Add tests
- [ ] Add support for remove from-prefix-requests
- [ ] Add README
- [ ] Add static methods for prefix requests (API)

### DIFF
--- a/nso-nipap/nso-nipap/src/java/src/net/spritelink/nsonipap/ConfigCdbSub.java
+++ b/nso-nipap/nso-nipap/src/java/src/net/spritelink/nsonipap/ConfigCdbSub.java
@@ -3,7 +3,6 @@ package net.spritelink.nsonipap;
 import net.spritelink.nsonipap.namespaces.*;
 
 import java.util.*;
-import java.math.BigInteger;
 
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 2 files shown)


nso-nipap/nso-nipap/src/java/src/net/spritelink/nsonipap/ConfigCdbSub.java
```diff
@@ -79,6 +77,246 @@ public void init() {
...
+     * @throws Exception
+     */
+    protected void addPrefixFromPrefix(String path, Prefix parentPrefix) throws Exception {
+
+        LOGGER.info("Create, From prefix request, path = " + path);
+        Prefix child_prefix = new Prefix();
+        try {
+
+            AddPrefixOptions child_opts = new AddPrefixOptions();
+            child_opts.put("prefix_length", "32");
```
garberg: We should probably not hard-code prefix lengths to 32.

nso-nipap/nso-nipap/src/yang/nipap.yang
```diff
@@ -14,6 +14,70 @@ module nipap {
...
+
+  grouping response-grouping {
+    container response {
+      description "CDB Subscriber will write the response here";
+      config false;
+      tailf:cdb-oper {
+        tailf:persistent true;
+        }
+
+      choice response-choice {
```
plajjan: This looks like somewhat of an anti-pattern to me. Better have a leaf for status, it can be an enum or something with "ok", "error" or whatever. Then you have a separate leaf that contains the error message. Right now you don't even have a leaf under the "ok" case so there's nothing there to read which is über weird ;) because you can't go to one single path and get the status, you have to probe around

PR #1059, opened 2016-09-27
merge tests from sql.py into nipaptest.py

### LINKED ISSUE #401: Creating an assignment that is 100% of a reservation fails
Hi,

I tried creating an assignment, say 192.168.0.0/24 from a reservation with the same prefix.
This fails.

By hand in postgres I get:

ERROR:  duplicate key value violates unique constraint "ip_net_plan__vrf_id_prefix__index"

Shall I try using the type in the unique constraint?

Kai

giganteous:
I have no personal travis, with these changes the tests seem to pass against master.

This is so I can add more tests later to fix #401.


giganteous:
It seems most tests are inside nipaptest.py. Perhaps the file sql.py should be dropped as a whole? or are there specific tests in there that might need to be moved?

plajjan:
@giganteous please take care of the indentation errors pointed out by @houndci-bot - i.e. non-multiple of four.

giganteous:
Thanks for the nudge. I suppose the noisy reports about long lines are.... somewhat noisy? If you have suggestions for those, I might be able to fix it.

plajjan:
LGTM

PR #1064, opened 2016-09-28
Match inherited tags

### LINKED ISSUE #1055: Tag searching no longer matches inherited tags
We recently upgraded from 0.28.4 to 0.29.3.  When searching using the web interface or pynipap's Prefix.smart_search, the tag search (string prefixed by #) no longer matches prefixes that have inherited the tag.

garberg:
Update the smart parser to match tags against a prefix's inherited as well as "own" tags.

Fixes #1055.


garberg:
Accidentally I also found that the creation of the citext extension never had worked in the SQL Makefile. Apparently the tests never covered it before :smile:

Issue #1074, opened 2016-10-11
Support newer Java release


garberg:
Currently the jnipap classes are built with java release 1.4 for various reasons. Fix support for newer releases (at least 1.7?).

PR #1075, opened 2016-10-11
Place exception classes in individual files

garberg:
Split up the jnipap exceptions into individual files. This fixes the warning
```
warning: auxiliary class ... in ... should not be accessed from outside its own source file
```
which is emitted when building the test cases.

part of #1074.


plajjan:
I fixed some small spelling error. Rest looks good to me, or as good as something can look when you don't understand how it works ;)

PR #1092, opened 2016-10-31
Replace empty string with null

### LINKED ISSUE #1085: Removing a vlan results in "Unhandled data error." in the web interface
Removing a vlan from a host by editing the prefix and clearing out the vlan field, then hitting "Save prefix" results in "Error Unhandled data error." dialog box.

version 0.29.5

garberg:
For integers, rewrite empty string with null values and strip whitespace. For strings these type of actions are performed server-side, but as the long-term goal is to move functionality from server to client-side I chose to put this code there.

Fixes #1085


plajjan:
LGTM

PR #1093, opened 2016-11-02
Added example import script for IPplan

### LINKED ISSUE #452: migration from IPPLAN to NIPAP
Helping people to take the final step of implementing NIPAP and moving all the data from IPplan.

Idea is maybe to get the info via IPplan API? 
First for testing and then when you know it works as intended, then just turn off the IPPLAN server, and live happily ever after with NIPAP. 
(link to api: http://iptrack.sourceforge.net/doku.php?id=ipplanapi )

amelialindsjo:
Fixes #452


utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+    Put your host information in the prefix object.
+    """
+    p = new_prefix()
+    p.prefix = str(host['ipaddr'])
+    p.type = "host"
+    p.description = host['description']
+    p.node = host['fqdn']
+    p.comment = ""
+
+    # Use remaining data from ipplan to populate comment field.
```
plajjan: Use AVP instead!?

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+# use local pynipap, useful if we are developing
+import sys
+sys.path.append('../pynipap')
+import pynipap
+import csv
+import ipaddress
+
+
+def add_ip_to_net(networks, host):
+    """
```
plajjan: Join this line with the next to follow PEP8 / PEP257 style guide on docstrings (https://www.python.org/dev/peps/pep-0008/#documentation-strings / https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).. and do so for the rest of the file :)

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+    networks = []
+
+    base = open(base_file, 'r')
+
+    csv_reader = csv.reader(base.readlines(), delimiter='\t')
+    for row in csv_reader:
+
+        # Rows with less than 3 entries is most likely a newline.
+        if len(row) < 3:
```
plajjan: "likely"? :P 

how many types of lines are there? If you are looking for a newline I think it should be looked for more explicitly. If you want to also check that we have a minimum of X rows then do that but since it's really an unhandled case I think you should be explicit and throw an error rather than silently ignoring it.

Something like

```python
if len(row) == 1 and row[0] == '\n': # ignore newlines
    continue
if len(row) < 3: # not enough data, weird line -> throw error
    raise ValueError("Unhandled line...")
```

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
+"""
```
plajjan: missing `#!/usr/bin/env python`

plajjan: or actually, seeing that you use ipaddress I guess this is py3 only, right? so `python3` :)

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+    csv_reader = csv.reader(base.readlines(), delimiter='\t')
+    for row in csv_reader:
+
+        # Rows with less than 3 entries is most likely a newline.
+        if len(row) < 3:
+            continue
+
+        network = {
+            'network': ipaddress.ip_network("{}/{}".format(row[0], row[2])),
+            'comment': row[1],
```
plajjan: Shouldn't this be description?

utilities/ipplan-import.py
plajjan: i.e. create an AVP for each attribute, like phone, mac and so forth.. don't cram it all together.

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Use remaining data from ipplan to populate comment field.
+    if 'additional' in host:
+        p.comment += "{}\n".format(host['additional'])
+
+    if len(host['location']) > 0:
+        p.comment += "Location: {}\n".format(host['location'])
+
+    if len(host['mac']) > 0:
+        p.comment += "MAC: {}".format(host['mac'])
```
plajjan: See above comment about not craming things together, but if that's really what you want to do (another way would be to have a command line argument for it).. then I think you are missing a \n here or is it intentional that MAC, Phone and User ends up on the same line?

utilities/ipplan-import.py
```diff
@@ -0,0 +1,181 @@
...
+            'hosts': []
+        }
+
+        networks.append(network)
+
+    base.close()
+
+    ipaddr = open(ipaddr_file, 'r')
+
+    csv_reader = csv.reader(ipaddr.readlines(), delimiter='\t')
```
plajjan: I don't know csv_reader by heart but do you really need to do readlines() here? Isn't enough with just feeding it the file handle and it will call __next__() on it? That way you don't load the entire file into memory.

utilities/ipplan-import.py
```diff
@@ -0,0 +1,187 @@
...
+            log.write("INFO: prefix: {}, type: {}, comment: {}\n".format(
+                p.prefix, p.type, p.description))
+
+        for host in network['hosts']:
+            p = add_host(host)
+            try:
+                p.save()
+            except Exception as exc:
+                log.write("ERROR: {}\n".format(exc))
+                log.write("INFO: host: {}, type: {}, node: {}, desc: {} \
```
plajjan: Sometimes I think Hound is over-zealous. I would recommend that you not break the text field of this line. Having the entire text field on one line makes it easier to grep for.

plajjan:
I think this looks good. The only thing I can think of is if it would be possible to include some example ipplan text files. I imagine it would make life easier for future maintenance work on this script so if you happen to have some ipplan data you can share, feel free to include.. if not, no biggie.

utilities/ipplan-import.py
```diff
@@ -0,0 +1,197 @@
...
+    base = open(base_file, 'r')
+
+    csv_reader = csv.reader(base, delimiter='\t')
+
+    buffer = ""
+    for row in csv_reader:
+
+        # Fixes quotation bug in ipplan exporter for networks
+        if len(networks) > 0 and len(buffer) > 0:
+            networks[-1]['comment'] += " ".join(buffer)
```
amelialindsjo: No, the script handles multi-line comments when they are quoted correctly. However there is a bug in the export script in ipplan that does not quote multi-line comments correctly for the base.txt file.

Here's the bug: https://github.com/hetznerZA/ipplan/blob/master/admin/exportbase.php#L75

PR #1105, opened 2016-11-25
nipapd: Fix override of configuration params

### LINKED ISSUE #1044: Syslog not working!?
Is syslog really working? I think I've seen multiple occurrences of where no log messages are written to the log...

garberg:
With the change from optparse to argparse, the override of configuration parameters from the config file with CLI arguments stopped working. This was due to argparse assigning the default value "False" to arguments with action store_true, while optparse assigned None. This was fixed by explicitly setting the default value to None.

Also did some cleanup of the code where the override is performed.

Fixes #1044


plajjan:
LGTM

PR #1106, opened 2016-11-25
Fix list_prefix with family arg

### LINKED ISSUE #1091: Unable to list prefixes filtered only by family
We are trying to use `list_prefix` to list all ipv6 prefixes in nipap. Family is the only attribute we are searching by. 

The response we get back is this:

> Unable to execute query: syntax error at or near "AND"
> LINE 36: ..._net_pool pool ON (inp.pool_id = pool.id)  WHERE  AND family...

Looking at the code, here:

https://github.com/SpriteLink/NIPAP/blob/bdae57d50bda4d9fc9dafb5c0a9ed273f9cc54a0/nipap/nipap/backend.py#L2382

It checks if `params` is empty but it never will be, as family is added to it on the line above it.

plajjan:
Heh, can't check len(params) == 0 when we set it on the line above,
guaranteeing it will always be at least 1...

Fixes #1091.

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -2378,11 +2378,11 @@ class Nipap:
         # prefix family needs to be handled separately as it's not stored
         # explicitly in the database
         if family:
-            params['family'] = family
             if len(params) == 0:
                 where = "family(" + prefix + "prefix) = %(family)s"
             else:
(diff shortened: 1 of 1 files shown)


garberg:
LGTM :+1:

PR #1111, opened 2016-11-30
Only show add-prefix-from-prefix option when it's usable

### LINKED ISSUE #863: "add prefix from prefix" option should not be accessible for prefixes of type host or assignment of max prefix length in the web interface

garberg:
Don't show the "Add prefix from prefix"-option in the prefix list popup menu when clicking prefixes with max prefix length, as no prefixes can be allocated from these.

Fixes #863


plajjan:
LGTM

PR #1117, opened 2016-12-19
Adding audit-search.py to the utilities directory.

Eising:
I've written a little script to work with the audit log. It's nothing fancy or anything, but it works.

The audit-search.py allows you to search through the audit log to see each change performed.

### DIFF
--- a/utilities/audit-search.py
+++ b/utilities/audit-search.py
@@ -1,0 +1,190 @@
+#!/usr/bin/env python
+""" NIPAP Audit searcher
+    ====================
+    This is a quick and dirty script to search the audit log db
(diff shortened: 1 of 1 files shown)


utilities/audit-search.py
```diff
@@ -0,0 +1,190 @@
...
+            output += "\n"
+
+        return output
+
+
+
+a = AuditLog()
+
+if len(sys.argv) == 1:
+    print "Use with prefix to search in audit log"
```
plajjan: We try to use py3 syntax so that print would be `print()`. If you still want to use py2 then you can do `from __future__ import print_function` to just get the new print function.

Issue #755, opened 2015-05-04
nipap-passwd --test-user fails for different default_backend


CrackerJackMack:
In nipap.conf if you set the default_backend to something other than local, then nipap-passwd fails when testing users with `--test-user`.  Same result when using two different sqlite auth sections.  Using the -f flag appears to have no affect.

```
[klandreth@localhost ~]$ nipap-passwd --test-user test -p test -f /etc/nipap/local_auth.db
The password seems to be wrong
[klandreth@localhost ~]$ ls -l /etc/nipap/local_auth.db
-rw-r--r-- 1 root root 3072 May  3 19:25 /etc/nipap/local_auth.db
```

This appears to only be the CLI tool nipap-passwd as the RPC and web-ui work fine despite testing failing.

plajjan:
Indeed, nipap-passwd doesn't do the right thing. https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap-passwd#L48 for example sets the db-file arguments (-f) to overwrite the config value for auth.backend.local.

There is a dependency on having a configuration file which we should probably try to decrease. I guess nipap-passwd would die completely if it can't find a config file at all, as per https://github.com/SpriteLink/NIPAP/blob/master/nipap/nipap-passwd#L45

Let's not try to read the config file at all if the auth_db file is specified with -f and when reading a configuration file we should probably iterate over the auth backends and try to find the local auth one.. if there is more than one we should throw an error. Instead the user would have to explicitly specify which auth_db file to work with.

ping-the-one:
Hi,

I also stumbled upon this but I have ldap as my default_auth, but have added a nipap-api user to the local_db.
When I run nipap-passwd test-user -u nipap-api@local -p kaka it says this:
`The user nipap-api@local does not exist`

When I run without "@" nipap-passwd test-user -u nipap-api -p kaka it throws an exception:

Traceback (most recent call last):
  File "/usr/sbin/nipap-passwd", line 84, in <module>
    if not auth.authenticate():
  File "/usr/lib/python2.7/dist-packages/nipap/authlib.py", line 372, in authenticate
    raise AuthError('Could not connect to LDAP server')

So I went on a tour in code and found this on row 79-81 in nipap-passwd:

       if not a.get_user(args.user):
            print "The user %s does not exist" % args.user
            sys.exit(2)


If you delete this block of code you will let authlib.py "choose" auth_backend by using @auth_backend_name.

Look at the these two rows in authlib.py
[...]

PR #1119, opened 2017-01-04
Fix for issue:#755 

ping-the-one:
Small change that makes it possible to use @ in the --username option argument to change auth_backend.
This only affect test-user.


plajjan:
Although we loose some detail in the error message, since we now can't differentiate between incorrect password and non-existing username, I think this is OK. We need to support multiple auth backends so this is taking us in the right direction.

PR #1124, opened 2017-01-25
"from-prefix" docstring fix

vazic:
from-prefix consumes list of prefixes rather than single prefix. Docstring does not reflect this

### DIFF
--- a/nipap/nipap/backend.py
+++ b/nipap/nipap/backend.py
@@ -2870,14 +2870,14 @@ class Nipap:
             From a prefix
                 Instead of specifying a pool, a prefix which will be searched
                 for new prefixes can be specified. In `args`, the key
-                :attr:`from-prefix` is set to the prefix you want to allocate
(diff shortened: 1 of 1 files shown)


nipap/nipap/backend.py
```diff
@@ -2870,14 +2870,14 @@ def find_free_prefix(self, auth, vrf, args):
             From a prefix
                 Instead of specifying a pool, a prefix which will be searched
                 for new prefixes can be specified. In `args`, the key
-                :attr:`from-prefix` is set to the prefix you want to allocate
-                from and the key :attr:`prefix_length` is set to the wanted prefix
-                length.
+                :attr:`from-prefix` is set to list of prefixes you want to
+                allocate from and the key :attr:`prefix_length` is set to
```
vazic: wrapped text a bit

plajjan:
thx! :D

PR #1131, opened 2017-02-16
Fix handling of monitor-attribute

### LINKED ISSUE #1130: Impossible to change the monitor flag from the web UI
It's not possible to change the prefix monitor flag from the web UI.

garberg:
Since the change from sending data from the client to the server side of the web UI with a POST containing JSON data, the handling of the monitor attribute has been broken in the web UI. The broken behaviour lead to it not only being impossible to change the monitor attribute from the web UI, but that it was disabled whenever updating a prefix in the web UI.

Fixes #1130


plajjan:
LGTM

PR #1132, opened 2017-02-18
Re-enable parsing of query string in URL

### LINKED ISSUE #1089: Search string lost when going"back" after editing a prefix
Some time ago the functionality which would make sure a specific search string is saved as a "state" in the browser history was lost. This means that if a user performs a search, edits a prefix and then clicks "back", they're met with a clean prefix search page, when he previously got the last performed search.

### LINKED ISSUE #1115: Saved search URLs are broken since upgrade from 0.28.4 to 0.29
In the former version (nipap-www 0.28.4) when we did a search on the web UI and saved its link, it worked again after opening it in another browser window. This behaviour changed after the upgrade. I think that this can be caused by the modification introduced in this changeset:

https://github.com/SpriteLink/NIPAP/blob/57834938ba84587ceaa07ac4ec901139890add70/nipap-www/nipapwww/public/nipap.js

Regarding to the git log message, the only change is that an extra slash is inserted before query_string, but if we modify the formerly saved

  https://ipam.company.com/prefix/list#query_string=stringToSearch&search_opt_parent=undefined&search_opt_child=undefined&explicit=true

style URLs to 

  https://ipam.company.com/prefix/list#/query_string=stringToSearch&search_opt_parent=undefined&search_opt_child=undefined&explicit=true

it simply omits the specified "stringToSearch" argument and lists all prefixes.

garberg:
Since the last upgrade of AngularJS, the part of the URL to the right of the hash has been prefixed with a slash. The library used to parse the URL considers the slash as part of the name of the parameter and thus the name needs to be updated when accessing it.

With this change, the URL is again successfully parsed.

This is a hack to re-enable the functionality lost with the AngularJS upgrade. This will all be replaced with an AngularJS implementation of the prefix list page.

Fixes #1115
Fixes #1089

### DIFF
--- a/nipap-www/nipapwww/templates/prefix_list.html
+++ b/nipap-www/nipapwww/templates/prefix_list.html
@@ -14,12 +14,12 @@
 	function handleUrlQuery() {
 
 		// Is there a query string in the URL?
-		if ($.url().fparam('query_string') == null) {
+		if ($.url().fparam('/query_string') == null) {
 			return true;
 		}
 
 		// query_string
-		$("#query_string").val(decodeURIComponent($.url().fparam('query_string')));
+		$("#query_string").val(decodeURIComponent($.url().fparam('/query_string')));
 
 		// search_opt_parent
 		if ($.url().fparam('search_opt_parent') != null) {


garberg:
This is a really hackish workaround. I just started working on an AngularJS implementation of the prefix list page which will render this obsolete, but it'd be great to have this workaround in place until it's ready to merge...

Issue #880, opened 2016-01-30
Move nipap-www.wsgi into /etc/nipap/www


plajjan:
See discussion in #796. This would allow us to only grant on the actual directory that holds the nipap-www.wsgi and not for all config files.

Don't forget to update docs/config-www.rst

PR #1139, opened 2017-03-16
Updated file path for paster invocation in docs

bbaja42:
Just a bug fix.

```
root@jessie:/home/vagrant/NIPAP# ls /etc/nipap/www
nipap-www.wsgi
```
```
root@jessie:/home/vagrant/NIPAP# paster serve /etc/nipap/nipap-www.ini
Starting server in PID 1689.
serving on 0.0.0.0:5000 view at http://127.0.0.1:5000
```

### DIFF
--- a/docs/config-www.rst
+++ b/docs/config-www.rst
@@ -37,10 +37,10 @@ Using paster is the easiest way to get the NIPAP web UI up and running, but
 it's not really suitable for production deployment. To serve the NIPAP web UI
 from paster, run the following::
 
-    paster serve /etc/nipap/www/nipap-www.ini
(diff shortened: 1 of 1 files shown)


plajjan:
Current one should be correct, also see https://github.com/SpriteLink/NIPAP/pull/1052 and https://github.com/SpriteLink/NIPAP/pull/1090 where this was last changed. Are you running the latest version?

The move was due to #880 to improve security / make it simpler.

We might have missed some location but the change was intentional so we will not move back, thus closing this :)

bbaja42:
Oh , you are right, i was running older version

PR #1140, opened 2017-03-29
Sanitize server response to UI code, using cgi.escape

bbaja42:
Jinja is already doing most of sanitization, but since % raw syntax is used at few places, this sanitization still helps with preventing XSS attack.

Helps with solving https://github.com/SpriteLink/NIPAP/issues/937
`


bbaja42:
To give example of existing XSS, and to verify this solves it:
 Go to http://nipap-demo.spritelink.net/, edit some prefix, and set following description

<script> alert("This is XSS :) " ) </script>

after this fix, this will not be possible, I think :)

bbaja42:
Hey guys, any thoughts on this change?

Issue #1116, opened 2016-12-13
nipap-cli doesn't see the 'global' section in .nipaprc file while parsing config


maxymax68:
nipap-cli doesn't see the 'global' section in .nipaprc file while parsing config

nipap address list FOO

Traceback (most recent call last):
  File "/usr/lib/python3.5/configparser.py", line 1135, in _unify_values
    sectiondict = self._sections[section]
KeyError: 'global'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/nipap", line 124, in <module>
    cmd.exe(cmd.arg, cmd.exe_options, args)
  File "/usr/lib/python3/dist-packages/nipap_cli/nipap_cli.py", line 507, in list_prefix
    v = get_vrf(opts.get('vrf_rt'), default_var='default_list_vrf_rt', abort=True)
  File "/usr/lib/python3/dist-packages/nipap_cli/nipap_cli.py", line 128, in get_vrf
    vrf_rt = cfg.get('global', default_var)
  File "/usr/lib/python3.5/configparser.py", line 778, in get
    d = self._unify_values(section, vars)
  File "/usr/lib/python3.5/configparser.py", line 1138, in _unify_values
    raise NoSectionError(section)
configparser.NoSectionError: No section: 'global'

plajjan:
Difficult to troubleshoot without seeing the config file - please post your .nipaprc (just censor passwords) :)

maxymax68:
Hi Kristian,

Here is the contents of my .nipaprc file:

# --------------------------------------
# NIPAP shell command configuration file
# --------------------------------------
#
# This file contains configuration for the NIPAP shell command. The 'nipap'
# executable will look for its config in ~/.nipaprc, so to successfully
# configure NIPAP, this file should be palced there and modified for your
# setup.
#

[global]

# NIPAP connection settings
hostname = 127.0.0.1
port = 1337
username = nipap-www
password = xxxxxxx

# Specify a command which can be used to complete node hostnames. This will be
# called when trying to complete a node hostname.
#
# The string "%search_string%" will be replaced by the current search string.
#complete_node_cmd = node list %search_string%

# Used as default VRF RT by 'nipap address add/modify/remove/view' commands
# when no other VRF RT is specified on the command line. Also used as default
# VRF when expanding a pool with a new prefix for the first time.
# '-' means VRF 'Default' (RT: -)
default_vrf_rt = -

[...]

PR #1143, opened 2017-03-30
Give nice error message if nipaprc isn't readable by the user

bbaja42:
Should help with https://github.com/SpriteLink/NIPAP/issues/1116

Example error
`
(nipap)➜  nipap-cli git:(barisa/cli) ✗ ./nipap
Failed to read file, make sure file /Users/barisa/.nipaprc is readable by current user
`

### DIFF
--- a/nipap-cli/nipap
+++ b/nipap-cli/nipap
@@ -68,6 +68,13 @@ if __name__ == '__main__':
         print("HINT: chmod 0600 ~/.nipaprc", file=sys.stderr)
         sys.exit(1)
 
+    try:
+        with(open(userrcfile)):
+            pass
+    except IOError as e:
+        print("Failed to read file %s, make sure file is readable by current user" % userrcfile, file=sys.stderr)
+        sys.exit(1)
+
     # config defaults
     cfg_defaults = {
         'hostname': 'localhost',
@@ -134,4 +141,3 @@ if __name__ == '__main__':
     except NipapError as exc:
         print("Command failed: %s" % str(exc), file=sys.stderr)
         sys.exit(1)
-


plajjan:
I don't immediately see what errors this will catch that aren't already caught by the code at https://github.com/bbaja42/NIPAP/blob/711f9e728f398e15c4842bd856536fd529ff258f/nipap-cli/nipap#L58-L69

Can you elaborate a little bit on when this code will help?

Do you need with `with` part?
I.e., isn't this
```
    try:
        with(open(userrcfile)):
            pass
```
better written as
```
    try:
        open(userrcfile)
```

bbaja42:
Well, with is required so that file handle is closed after opening it.
But since this is a CLI script, no harm is done even if the file handle is
left open; and as result the code is smaller.

If you'd like, I can remove it.

plajjan:
Bah, I just saw your comment about #1116 😜

plajjan:
Hmm, I would assume that there is no open file handle because you don't catch the return value of open() so python gc will immediately see it's not used and close it again but this is an assumption, perhaps an incorrect one, from my side.

bbaja42:
I think you are right about GC destroying the variable, but the file handle needs manual closing afaik.
http://preshing.com/20110920/the-python-with-statement-by-example/

plajjan:
Yeah, you're right. File descriptor is still there. I checked in /proc/X/fd for python3 REPL open()

plajjan:
Anyway, this looks good so I'm merging. Thanks! 😄

bbaja42:
Wohoo :)

PR #1153, opened 2017-04-24
Fix typo in password example in documentation

bhirbec:
Hi,

There's typo [here in the documentation](http://spritelink.github.io/NIPAP/docs/config-www.html). The password is "s3cr3t" in the first command and it becomes "s3cret" in the second one. 

Thanks,
Ben


plajjan:
LGTM :)

PR #1180, opened 2018-01-17
Fix path for AuthOptions

plajjan:
The example shows importing pynipap and directly importing some names like VRF,
Prefix and Pool but AuthOption is not among them so we would need to refer to it
as pynipap.AuthOption. It's important we are consistent in our documentation.


garberg:
LGTM :+1:

garberg:
Friendly reminder :smile:

PR #1195, opened 2018-07-03
Use sync bind for searching, needed by MS AD

### LINKED ISSUE #1193: Use sync bind for searching, needed by MS AD
With my corp MS AD, I get:
00002024: LdapErr: DSID-0C0605D4, comment: No other operations may be performed on the connection while a bind is outstanding., data 0, v1db1
without the '_s', and it works with it - if this is intentionally async, then it should be configurable I think?

PS. complete newbie in python/ldap/nipap, so please don't shoot me :/

steffenschumacher:
With my corp MS AD, I get:
00002024: LdapErr: DSID-0C0605D4, comment: No other operations may be performed on the connection while a bind is outstanding., data 0, v1db1
without the '_s', and it works with it - if this is intentionally async, then it should be configurable I think?

Fixes #1193

PS. complete newbie in python/ldap/nipap, so please don't shoot me :/

### DIFF
--- a/nipap/nipap/authlib.py
+++ b/nipap/nipap/authlib.py
@@ -387,7 +387,8 @@ class LdapAuth(BaseAuth):
         try:
             # Create separate connection for search?
             if self._ldap_search_conn is not None:
-                self._ldap_search_conn.simple_bind(self._ldap_search_binddn, self._ldap_search_password)
+                self._ldap_search_conn.simple_bind_s(
+                    self._ldap_search_binddn, self._ldap_search_password)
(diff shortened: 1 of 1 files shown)


webminster:
Why has this never been merged (or similar merged)?   I had this exact issue in current codebase, and had to change it locally to work.
-Alan

steffenschumacher:
I think this project is next to abandoned I’m afraid.
I did actually nearly complete a port to py3.6, but never completed it due to lack of time, but since the authors aren’t active anymore I’m migrating to something else soonish..

PR #1204, opened 2018-11-26
Enable LDAP module

damluji:
Enable LDAP(S) and install prerequisites.
fix logging, 
try to improve authentication in a better way for use cases that cannot authenticate with a '@' symbol.


garberg:
Thanks for contributing!

To understand the proposed changes, what problem did you want to address by adding the secondary_backend? Wouldn't you achieve the same thing by simply making ldap1 the default backend (which would be used for usernames without "@" in them), and using the username "localadmin@local" when you want to authenticate using the local auth backend? Hard-coding usernames in the auth library is something I'd really like to avoid.

PR #1211, opened 2019-01-29
Fixed removal of extra attributes

### LINKED ISSUE #1205: nipap-www: Attribute removal in wrong order
Ive noticed what appears to be incorrect behaviour when removing attributes in the web UI.

Say I have two attributes: a=a and b=b

They are ordered in the web UI as:

a=a
b=b

To the right of each attribute is a remove button. Suppose I click the remove button for attribute "a", "b" actually disappears from the list!

garberg:
Fixes a bug where the last extra attribute added to a prefix would be removed, no matter which one was clicked.

Fixes #1205


plajjan:
LGTM. Merging.

PR #1213, opened 2019-02-22
Increase search delay to 500 ms

### LINKED ISSUE #1208: Search is too quick when typing
This is a badly described issue. 

Search is trying to find what users type too quick and ends up prolonging the search. Most of the time our users know what equipment they want to look at and now write the name somewhere else and then paste the full name as a workaround as typing ends up being slower.

A way to disable autocomplete?

garberg:
Users have reported the delay from typing a character to a search being performed is too short. Increasing from 200 to 500 ms.

Fixes #1208


plajjan:
LGTM. Merging.

PR #1239, opened 2020-02-03
Updated Code-Based Directions

jpartain89:
I added the `sudo` commands to the copy-pastable parts of the directions, just to make it a little easier for people to install everything. Sometimes, some new people to linux might not yet have the `sudo` requirement linked in their heads, so this just removes that barrier.


garberg:
Nice, thanks for contributing!

PR #1243, opened 2020-03-30
Changed syntax compatible to Python 3

ChrisRo89:


nipap/setup.py
```diff
@@ -14,7 +14,7 @@ def get_data_files():
         subprocess.call(["rst2man", "nipapd.man.rst", "nipapd.8"])
         subprocess.call(["rst2man", "nipap-passwd.man.rst", "nipap-passwd.1"])
     except OSError as exc:
-        print >> sys.stderr, "rst2man failed to run:", str(exc)
+        print("rst2man failed to run:", exc)
```
emjemj, line 17: You should add file=sys.stderr to print call to avoid printing error messages to stdout.

PR #1244, opened 2020-05-01
nipap/xmlrpc: update legacy flask extension import

chrised:
Updated the flask.ext.compress import to the new underscore format.

## Reason
This actually broke on an ancient Ubuntu xenial server I upgraded recently.
I had to edit `xmlrpc.py` and insert the correct import.

Towards #946


emjemj:
Would be nice with a try/except to be backwards compatible with the old library.

garberg:
Implemented in #1273 and #1276.

PR #1251, opened 2020-09-08
Update triggers.plsql

BjornPerssonSigma:
Fixed pool constraint so that two prefixes in different VRFs are not allowed in a pool.


garberg:
LGTM! Some bookkeeping is required for the automated upgrades etc to function, but I can take care of that as it's not that well documented ATM.

Issue #1254, opened 2020-09-09
port nipap-www to python 3.6


steffenschumacher:
Update syntax, libs etc to enable running on ubuntu:bionic

garberg:
Fixed in #1316

Issue #1252, opened 2020-09-08
port nipapd to python 3.6


steffenschumacher:
In order to preserve this great app, I'd like to pitch in with some refurbishing of nipapd - I might do nipapwww also eventually..

* 2to3:
All the classics: print, exceptions, instanceof, 0<=mask<=32 support, remove unicode() etc etc
* optimize string formatting:
Py3 brings more readable str.format, but as performance is key, I'm going to only use it for exceptions.
Concatenation is by far the most performant for strings, so this is preferred and otherwise the % operator is used
For logging, %s allows params to be embedded only if the logging is not filtered out, so this is implemented to avoid unneeded execution of the % operator.
I'll harmonise a few things like consistently using " around ' and vice versa, and never \' or using ''' around '..
The way strings are being formatted has a big impact when done often, and since sql is generated often it will pay to make it more efficient.
* update libs / replace libs:
ConfigParser -> configparser
psycopg2 -> psycopg2-binary (and drop use of Inet class in backend)
flask-xml-rpc -> flask-xml-rpc-re (former is dead since 2010 and py2 only)
updated flask-compress import syntax

PR #1257, opened 2020-09-13
#1254: Update syntax, libs etc to enable running on ubuntu:bionic

steffenschumacher:
NOTE! requires nipapd to be py36 capable also, so this will not work
independently - see PR for #1252


steffenschumacher:
Actually hold off on this one - I somehow missed that pylons is py2.x only, so I need to port the whole thing to pyramid, which will take me quite some time, since I have next to no experience with frontend, pylons let alone pyramid.. I'll see what I can do.

PR #1267, opened 2020-11-16
Fixing slow load on VRF page and adding functionality for infinite sc…

BjornPerssonSigma:
Fix on the VRF page so that users can search before the loading of all data has been made. Have also implemented infinite scroll, that is loading data by scrolling, so that it fits with angular.js.

The change has shortened the loading time and made it possible to search before the initial load has been made. There are still some delay of around 4-6 seconds before the initial load will show on the screen.

License for Infinite scroll has been added to the ng-infinite-scroll.min.js - file.

### DIFF
--- a/nipap-www/nipapwww/public/controllers.js
+++ b/nipap-www/nipapwww/public/controllers.js
@@ -9,21 +9,119 @@ var nipapAppControllers = angular.module('nipapApp.controllers', []);
  */
 nipapAppControllers.controller('VRFListController', function ($scope, $http) {
 
-	// Fetch VRFs from backend
-	$http.post('/xhr/list_vrf',
-		JSON.stringify({}),
-		{ 'headers': { 'Content-Type': 'application/json' } })
-		.then(function (response) {
-			if (response.data.hasOwnProperty('error')) {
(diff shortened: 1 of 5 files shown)


nipap-www/nipapwww/public/controllers.js
```diff
@@ -9,21 +9,119 @@ var nipapAppControllers = angular.module('nipapApp.controllers', []);
...
-				showDialogNotice('Error', response.data.message);
-			} else {
-				$scope.vrfs = response.data;
-			}
-		})
-		.catch(function (response) {
-			var msg = response.data || "Unknown failure";
-			showDialogNotice('Error', response.status + ': ' + msg);
-		});
+	var isSmartSearch;
```
BjornPerssonSigma, line 12: Can't see this extra indent in my code. It should be a new blank row below "nipapAppControllers.controller('VRFListController', function ($scope, $http) {"

image

I tried to follow how the following method was written:
 "/*
 * PoolListController - used to list pools on the /pool/list-page
 */
nipapAppControllers.controller('PoolListController', function ($scope, $http) {"
image

garberg:
LGTM, thanks a lot!

PR #1269, opened 2020-12-03
fix(nipap-cli): Allow `None` as ConfigParser values

jvalcke:
It seems that newer python versions (I'm on python 3.8) do not accept
`None` by default. Adding the `allow_no_value` argument fixes that.


garberg:
Nice, thx for contributing!

PR #1274, opened 2021-03-11
[requires.io] dependency update on master branch

plajjan:


garberg:
requires.io seems dead (and replaced with dependabot). Closing.

PR #1281, opened 2021-03-18
Rest API feature

BjornPerssonSigma:
### Change in this Pull request

Adding some REST functionality to NIPAP.

The change will make it possible to do XMLRPC and REST requests to the same IP and Port.
BUT, as of now, all functions aren't set up in in the new functionality.

The authentication used in xmlrpc.py is used in rest.py, but to be able to use decorators correctly i had to recieve the request first and then send my args to a helper function where changes to args are made, see below:

```
@requires_auth
    def list_prefixes(self, args):
        return args

    
    def get(self):
        combined_args = self.combine_request_args(False)
        args = self.list_prefixes(combined_args)

        try:
            result = self.nip.list_prefix(args.get('auth'), args.get('prefix') or {})
            # mangle result
            for prefix in result:
                prefix = _mangle_prefix(prefix)

            result = jsonify(result)
            return result
        except (AuthError, NipapError) as exc:
            self.logger.debug(unicode(exc))
            abort(500, error={"code": 500, "message": exc})
```

I'm not very proud of the solution but it works, without doing major changes to the "requires_auth" decorator.

### Limitations
Per this pull request only the following NIPAP functions are provided by the REST API:

- list_prefix
- add_prefix
- remove_prefix

The above are done through GET, POST and DELETE to the endpoint /rest/v1/prefixes

PUT needs to be added for editing prefixes, it's not with this PR.

Further development should be made to add the other functionalities that are in the XMLRPC API.

### DIFF
--- a/nipap/debian/control
+++ b/nipap/debian/control
@@ -17,7 +17,7 @@ Description: Neat IP Address Planner
 
 Package: nipapd
 Architecture: all
-Depends: debconf, nipap-common, python (>= 2.7), ${misc:Depends}, python-psycopg2, python-flask, python-flask-xml-rpc, python-flask-compress, python-tornado, python-parsedatetime, python-tz, python-dateutil, python-psutil, python-pyparsing
+Depends: debconf, nipap-common, python (>= 2.7), ${misc:Depends}, python-psycopg2, python-flask, python-flask-xml-rpc, python-flask-restful, python-flask-compress, python-tornado, python-parsedatetime, python-tz, python-dateutil, python-psutil, python-pyparsing
 Description: Neat IP Address Planner XML-RPC daemon
  The Neat IP Address Planner, NIPAP, is a system built for efficiently managing
  large amounts of IP addresses. This is the XML-RPC daemon.
--- a/nipap/nipap/rest.py
+++ b/nipap/nipap/rest.py
@@ -1,0 +1,221 @@
+""" NIPAP REST
+    ==============
+    This module contains the actual functions presented over the REST API. All 
+    functions are quite thin and mostly wrap around the functionalty provided by
+    the backend module.
+"""
+
+import datetime
+import logging
+import time
+import pytz
+import json
+from functools import wraps
+from flask import Flask, request, Response, got_request_exception, jsonify
+from flask_restful import Resource, Api, abort
+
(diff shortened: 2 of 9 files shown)


nipap/debian/control
```diff
@@ -17,7 +17,7 @@ Description: Neat IP Address Planner
 
 Package: nipapd
 Architecture: all
-Depends: debconf, nipap-common, python (>= 2.7), ${misc:Depends}, python-psycopg2, python-flask, python-flask-xml-rpc, python-flask-compress, python-tornado, python-parsedatetime, python-tz, python-dateutil, python-psutil, python-pyparsing
+Depends: debconf, nipap-common, python (>= 2.7), ${misc:Depends}, python-psycopg2, python-flask, python-flask-xml-rpc, python-flask-restful, python-flask-compress, python-requests, python-tornado, python-parsedatetime, python-tz, python-dateutil, python-psutil, python-pyparsing
```
garberg, line 20: python-requests is only needed by the test suite, right? In that case, don't list it as a dependency for nipapd.

nipap/nipap/xmlrpc.py
```diff
@@ -24,8 +24,7 @@
 from authlib import AuthFactory, AuthError
 
 
-def setup():
-    app = Flask('nipap.xmlrpc')
+def setup(app):
     Compress(app)
```
garberg, line 28: Put this in nipapd, where app is created.

nipap/nipap/rest.py
```diff
@@ -0,0 +1,240 @@
...
+        elif request_queries and not isPost:
+            args.update({'prefix': request_queries})
+        
+        if request_body:
+            args.update(request_body)
+
+        return args
+
+
+if __name__ == '__main__':
```
garberg, line 239: This block we can remove. It won't work as it stands today and I don't see when we'd like to use it.

nipap/debian/control
BjornPerssonSigma: Right!
Change has been made in the newest commit.

nipap/nipap/xmlrpc.py
BjornPerssonSigma: Done in new commit.

PR #1282, opened 2021-03-25
Adding REST API feature

BjornPerssonSigma:
### Change in this Pull request
Adding some REST functionality to NIPAP.

The change will make it possible to do XMLRPC and REST requests to the same IP and Port.

### Limitations
Per this pull request only the following NIPAP functions are provided by the REST API:

- list_prefix
- add_prefix
- remove_prefix

The above are done through GET, POST and DELETE to the endpoint /rest/v1/prefixes

PUT needs to be added for editing prefixes, it's not with this PR.

Further development should be made to add the other functionalities that are in the XMLRPC API.
This should/might be done by adding more classes in rest.py


garberg:
Looks good, merging!

PR #1293, opened 2021-08-24
nipapd: improve smart_search_vrf search

martintopholm:
When NIPAP page rendering was slow we found that removing this sleep
wastly improved performance. It was introduced with commit 1da02003
(www: Improvements of VRF list smart search).


garberg:
LGTM, thanks!

PR #1295, opened 2021-09-13
Script to remove expired quarantined prefixes

MikaKeljaNexer:


MikaKeljaNexer:
First PR

utilities/remove_expired.py
```diff
@@ -0,0 +1,94 @@
...
+    Prefixes within prefixes are recursively deleted only if the
+    type is Assigned.
+"""
+
+import ConfigParser
+import os
+import datetime
+# use local pynipap, useful if we are developing
+import sys
+sys.path.append('../pynipap')
```
garberg, line 21: This can be removed

utilities/remove_expired.py
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    try:
+        res = Prefix.search(query)
+    
+    except socket.error:
+        print >> sys.stderr, "Connection refused, please check hostname and port"
+        sys.exit(1)
+    except xmlrpclib.ProtocolError:
```
garberg, line 49: Hm, won't an authentication error raise some NipapAuthError?

utilities/remove_expired.py
MikaKeljaNexer: Removed

utilities/remove_expired.py
MikaKeljaNexer: Yes, I changed to NipapAutError and added also a general exception.

MikaKeljaNexer:
Updated after review comments and also made the script to Python3

garberg:
LGTM, thanks for contributing!

PR #1299, opened 2021-10-20
The Expires date field was always empty in the UI because the datepic…

MikaKeljaNexer:
The UI bootstraP datepicker control expects that the date in the model is a Javascript Date object so it can handle it correctly.
If it is a string it cannot display the date.


garberg:
Thanks! CI fixed in master branch, please rebase.

PR #1312, opened 2022-11-05
REST API case-insensitive search - fixes

garberg:
Some fixes for #1310.

Added support for case insensitive search for string data values in the REST API + added some new tests for the REST API.
Also improved the error handling a bit so that the error message is returned correctly.


garberg:
LGTM :+1:

PR #1316, opened 2022-11-10
Move to Python 3

garberg:
It's way overdue, but a move to Python 3. Based on the work in #1253.


garberg:
As the authentication library is shared between nipapd and nipap-www, Python 3 support needs to be fixed for nipap-www as well before this can be merged. Unfortunately Pylons, the web framework used, lacks Python 3 support and is abandoned. To prepare for this, a migration from Pylons to Flask has been initiated.

PR #1329, opened 2023-07-04
Replace Pylons with Flask

garberg:
Replaced the web framework Pylons, which is abandoned since years, with Flask. Introducing Flask requires a new configuration setting in nipap.conf, but the Debian package should add that automatically if the user permits it so the upgrade should be seamless.


garberg:
Okay, so GitHub pulled the plug on action runners on Ubuntu 18.04 so the tests are not running. As there are really no tests to the web UI and I've done quite a bit of manual testing with regards to packaging etc, I'll merge this anyway. It's a step on the path to supporting Ubuntu > 18.04 >_<.

PR #1331, opened 2023-07-05
Run CI tests on Ubuntu 20.04

garberg:
As GitHub have phased out Ubuntu 18.04 images, change the runner image to Ubuntu 20.04.


garberg:
Fixed in #1316

PR #1337, opened 2023-10-06
Handle edit_prefix without VRF

garberg:
Fixed a bug in the backend where the edit_prefix operation would fail for prefixes in a VRF where the VRF wasn't included in the prefix attributes.


garberg:
LGTM - will check CI when rebased into py3-branch.

PR #1347, opened 2023-10-16
Handle binary data from LDAP

garberg:
It seems the Python 3 version of the LDAP module returns bytes instead of strings. This caused the full name fetched from LDAP, now a bytes object, to be passed around and sent to the backend as an xmlrpc.client.Binary from the web UI and causing a crash.

Updated the LdapAuth class to decode the received full name and encode strings to bytes before determining group membership.


garberg:
LGTM

PR #1355, opened 2023-11-01
Add OpenTelemetry tracing

erikoqvist:
Add OpenTelemetry tracing for cli, www and nipapd

### DIFF
--- a/pynipap/tracing.py
+++ b/pynipap/tracing.py
@@ -1,0 +1,111 @@
+import inspect
+import sys
+from functools import wraps
+from pynipap import AuthOptions
+
+if sys.version_info[0] < 3:
+    import xmlrpclib
+
+    int = long
+else:
+    import xmlrpc.client as xmlrpclib
+
+try:
+    from opentelemetry import trace, context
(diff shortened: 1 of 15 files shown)


pynipap/tracing.py
```diff
@@ -0,0 +1,110 @@
...
+                             use_builtin_types=use_builtin_types,
+                             headers=headers)
+
+        def request(self, host, handler, request_body, verbose=False):
+            with tracer.start_as_current_span("POST XML request", context.get_current(), SpanKind.CLIENT):
+                current_span = trace.get_current_span()
+                try:
+                    result = super().request(host, handler, request_body, verbose)
+                    current_span.set_attribute("http.status_code", 200)
+                except xmlrpclib.ProtocolError as protocolError:
```
garberg, line 90: Shouldn't we make sure to re-raise the exception, this'll silently drop all ProtocolErrors?

erikoqvist: Yes, we definitely should. I've pushed an update correcting this.

garberg:
LGTM :+1:

PR #1361, opened 2023-11-02
rest: Add pagination support

erikoqvist:


garberg:
Upgrade tests due to NIPAP not properly restarting after the upgrade. I'm working on that one... LGTM!

PR #1379, opened 2024-05-14
Add opentelemetry tracing sampler configuration

erikoqvist:


nipap-cli/nipap
```diff
@@ -89,6 +89,16 @@ if __name__ == '__main__':
     nipap_cli.nipap_cli.cfg = cfg
 
     if cfg.has_section("tracing"):
+
+        sampler = None
+
+        if cfg.has_option("tracing", "otel_traces_sampler"):
+            trace_sampler = cfg.get("tracing", "otel_traces_sampler")
+            from opentelemetry.sdk.trace.sampling import _KNOWN_SAMPLERS
```
garberg, line 97: This'll not fail gracefully if the opentelementry libs are not installed (as the rest of the implementation) - perhaps we should make this validation in tracing.py instead?

erikoqvist: Added try-except to print error if opentelemetry depencies are missing

garberg:
LGTM :+1:

PR #1396, opened 2024-08-14
Merge pynipap to single file

garberg:
To avoid hogging the global name "tracing", merged tracing.py into pynipap.py.

pynipap is a bit of a beast and really should be split up in smaller files in a python package, but I'll leave that for another day...


garberg:
Fixed in #1412 instead.

PR #1418, opened 2025-01-10
Run tests on Ubuntu 24.04

garberg:


garberg:
I think I've now hit the same error as when I last tried to update Flask... So that'll have to happen first. Depends on #1360.

PR #1419, opened 2025-01-13
Migrate pynipap to setuptools

garberg:
As distutils is being phased out, let's initiate the work with migrating to something more modern. Let's start with pynipap.


garberg:
There seems to be some bug in the build chain (pybuild, pyproject...) which is resolved in Ubuntu 24.04, thus the pipeline will need to be migrated to 24.04 before this can be merged.

PR #1420, opened 2025-01-13
WIP: Repackage nipapd

garberg:
Repackage nipapd.


garberg:
Some bug in the build tools causes this to fail on Ubuntu 22.04, but succeeds on 24.04. Thus this depends on #1418.

PR #1433, opened 2025-05-05
Fix typo when searching pools

luis-c-matos-alb:


plajjan:
@luis-c-matos-alb can you strip out the extraneous commits from this PR? I think that initial commit to fix the backend.py looks correct but the rest is automatic noise?

@garberg should there be a test that covers this? I guess we can merge in the meantime since the fix looks obvious

luis-c-matos-alb:
New pullRequest https://github.com/SpriteLink/NIPAP/pull/1434

garberg:
> @luis-c-matos-alb can you strip out the extraneous commits from this PR? I think that initial commit to fix the backend.py looks correct but the rest is automatic noise?
> 
> @garberg should there be a test that covers this? I guess we can merge in the meantime since the fix looks obvious

Agreed, we can fix the test later.

